PR #57, opened 2016-09-22
add gradient to line chart

rwholey-eb:


aganov:
Is there ability to change `singleLineGradientColors`, looking at the source code I can't find it?

Golodhros:
Hi @aganov, thanks for pointing that out!

There isn't any accessor for it, but there should be one. 

I created an issue for it https://github.com/eventbrite/britecharts/issues/234 
We'll tackle it soon!

Tobbe:
This confused me a lot. I was expecting single line graphs to honor the `.colorSchema` setting, alas it doesn't. Was just about to write an issue about it when I found this one.

So one idea would be to run the gradient through the colors passed to `.colorSchema`. What do you think about that?

PR #125, opened 2017-02-08
Rw expand x axis capabilities

rwholey-eb:
diff starts here https://github.com/eventbrite/britecharts/compare/master...ryanwholey:rw-expand-x-axis-capabilities#diff-1be418fa346b7410a50fd56a54f59ff4

we need to fix this build system :(


src/charts/line.js
```diff
@@ -835,7 +878,29 @@ define(function(require){
...
+        exports.forceAxisFormat = function(_x) {
+            if (!arguments.length) {
+              return forceAxisSettings || defaultAxisSettings;
+            }
+            forceAxisSettings = _x;
+            return this;
+        };
+
+        /**
+         * constants to be used to force the x axis to respect a certain granularity
```
miglesiasEB: Could we add here the possible values for this property?

rwholey-eb: yeah of course!

miglesiasEB:
LGTM

PR #133, opened 2017-02-10
Updating Stacked area chart to not use Set

Golodhros:
Due to a bug in EB core, we need to avoid the use of Set for the moment.


docs/scripts/bundle.js
```diff
@@ -3047,6 +3045,12 @@
 	    var ONE_AND_A_HALF_YEARS = 47304000000;
 	    var ONE_DAY = 86400001;
 	
+	    var uniq = function uniq(arrArg) {
```
rwholey-eb: why not use the underscore one

src/charts/stacked-area.js
```diff
@@ -21,6 +21,9 @@ define(function(require){
     const ONE_AND_A_HALF_YEARS = 47304000000;
     const ONE_DAY = 86400001;
 
+    const uniq = (arrArg) => arrArg.filter((elem, pos, arr) => arr.indexOf(elem) == pos);
```
rwholey-eb: meant to write that here

docs/scripts/bundle.js
miglesiasEB: Less dependencies!

PR #137, opened 2017-02-11
updated script logo

evb-sun:


docs/styles/custom.css
```diff
@@ -1897,8 +1897,8 @@ samp {
 code {
   padding: 2px 4px;
   font-size: 90%;
-  color: #c7254e;
```
miglesiasEB: Oh! Bummer!
This is the wrong file Sun! You need to do this on /src/doc/template/static/styles, as this file is being generated!

PR #139, opened 2017-02-12
Adding aspect ratio to stacked area and changing demo data

### LINKED ISSUE #94: Nicer data for Stacked Area Demo
We need to get a nicer dataset for the stacked area chart demo. The current set results in an almost psychedelic effect with all the colors, it should be something more smooth.

maybe use any of the data generation projects out there?

Golodhros:
Fixes #94 and adds the aspect ratio setter to the Stacked Area chart.
It also adds those tests to the line chart test suite.


rwholey-eb:
Looks good to me!

PR #140, opened 2017-02-12
Font and Logo

evb-sun:
@miglesiasEB  I've added Benton Light and Regular to Britecharts, but the theme on Bootswatch still uses Google fonts. Not sure how to import fonts over there.

Also I've updated changed Script Logo again..


Golodhros:
Great! Thanks Sun!

Yeah, I tried to add Benton in the theme and I couldn't figure out how, I will give it another try soon.

Issue #146, opened 2017-02-18
Deprecate the LineChartPointByDate data on the linechart


Golodhros:
The LineChartPointByDate is a grouping by date we are passing as an input to the line chart:
http://eventbrite.github.io/britecharts/global.html#lineChartPointByDate

We think this should be done by the linechart itself from the ungrouped data we input. We do something similar on the Stacked area chart.

PR #147, opened 2017-02-21
Deprecate the LineChartPointByDate data on the linechart #146

Golodhros:
We want to clean up the data input of the Line chart, making it not required to pass in a 'dataByDate'.

In this PR I have updated the API for the line, stacked-area and tooltip charts.

NOTE: This changes are breaking changes, as they change the structure of the data expected by the Line Chart. Please check the line chart data type to see the new structure.


docs/line.js.html
```diff
@@ -358,7 +317,7 @@ <h1 class="page-title">Source: line.js</h1>
         function exports(_selection) {
             _selection.each(function(_data) {
                 ({
-                    data,
+                    dataByTopic,
                     dataByDate
                 } = cleanData(_data));
```
rwholey-eb: so we still have to pass it a proeprty on _data that is dataByDate?

src/charts/line.js
```diff
@@ -458,20 +417,54 @@ define(function(require){
...
-        function cleanData({data, dataByDate}) {
-            data.forEach(function(kv) {
-                kv.Data.forEach(function(d) {
-                    d.date = new Date(d.fullDate);
+         * @param  {obj} dataByTopic    Raw data grouped by topic
+         * @return {obj}                Parsed data with dataByTopic and dataByDate
+         */
+        function cleanData({dataByTopic, dataByDate}) {
+
+            if (dataByTopic) {
```
rwholey-eb: ah i see its optional here

docs/line.js.html
```diff
@@ -580,20 +539,54 @@ <h1 class="page-title">Source: line.js</h1>
...
-         * @param  {obj} dataByDate     Raw data ordered by date
-         * @return {obj}            Parsed data with dates
+         * @param  {obj} dataByTopic    Raw data grouped by topic
+         * @return {obj}                Parsed data with dataByTopic and dataByDate
          */
-        function cleanData({data, dataByDate}) {
-            data.forEach(function(kv) {
-                kv.Data.forEach(function(d) {
-                    d.date = new Date(d.fullDate);
+        function cleanData({dataByTopic, dataByDate}) {
```
rwholey-eb: it would be interesting if we could think of a dynamic way to index the dataset by the key that you'd like, instead of having to pass it in. Not quite sure how it would play out yet, but it would be nice!

docs/line.js.html
miglesiasEB: Well, this is the output

PR #159, opened 2017-03-07
create format helpers

ryanwholey:
Change exponential label format to n-999, 1k-nk format

## Description
The tooltip format looked good, so i extracted it to a format helper, then used that format helper on the line chart y axis

## Motivation and Context
yaxis was showing exponential format instead of 1000s

## How Has This Been Tested?
common helper tested,
no test written for format helper in case we decide to change the formats

screen shot 2017-03-06 at 6 11 32 pm


miglesiasEB:
This is awesome!!

Could you add it to the Stacked Area Chart too? Just today Nahuel asked me about that same yAxis value format!!

Thanks!!!

ryanwholey:
oh fsho! i'll do that now, thanks!

ryanwholey:
image

miglesiasEB:
Thanks Ryan!

PR #165, opened 2017-03-15
add force order to tooltip

ryanwholey:
Provide an option to force the tooltip to display in any order you like

## Description
added api to the tooltip to pass an order array

## Motivation and Context
Other category wasn't at the bottom despite being passed at the bottom, think it makes our tooltip more usable to be able to define the order

## How Has This Been Tested?
api testing method, plugged in to core

image


demos/demo-line.js
```diff
@@ -90,7 +91,8 @@ function createLineChart(optionalColorSchema, optionalData) {
         chartTooltip
             // In order to change the date range on the tooltip title, uncomment this line
             // .forceDateRange(chartTooltip.axisTimeCombinations.HOUR_DAY)
-            .title('Quantity Sold');
+            .title('Quantity Sold')
+            .forceOrder(dataset.dataByTopic.map((topic) => topic.topic));
```
miglesiasEB: Look out! The demos code is not ES6!

ryanwholey: holy! for real! why not? we run it through webpack

PR #167, opened 2017-03-15
update color selection order for bar

ryanwholey:
Use color map to determine which color to make each horizontal bar

## Description
Have the bar charts use the color map defiened here to determine which color they should fill each bar.
Also, move green to the top of the britecharts color schema, per suns recommendation

## Motivation and Context
Bar chart colorScale was giving us some really weird responses, we want to make sure that the first bar always matches the first color in the color schema. 

## How Has This Been Tested?
All tests pass, visually looks good and in core also works like a charm

image

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -68,8 +68,8 @@ define(function(require) {
             chartWidth, chartHeight,
             xScale, yScale,
             colorSchema = colorHelper.colorSchemas.singleColorAloeGreen,
-            colorScale,
-            topicColorMap,
+            colorList,
+            colorMap,
             numOfVerticalTicks = 5,
             numOfHorizontalTicks = 5,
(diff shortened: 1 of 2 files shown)


src/charts/bar.js
```diff
@@ -207,13 +206,15 @@ define(function(require) {
...
                     .padding(0.1);
             }
-            colorScale = d3Scale.scaleOrdinal().range(colorSchema).domain(data.map((_, i) => i))
 
-            let range = colorScale.range();
-            topicColorMap = colorScale.domain().reduce((memo, item, i) => {
-                memo[item] = range[i];
-                return memo;
-            }, {});
+            colorList = data.map(d => d)
```
ryanwholey: ahh we need it reversed before we use the index in the function below, and we dont want to reverse the data so we need a map first

PR #219, opened 2017-05-12
Rw cdn bundle

ryanwholey:
We need a bundle that can be loaded via a script tag

## Description
I propose a bundle that can be loaded via script tag that doesnt mess with our current prod files. We should really take a look at reworking our webpack stuff but for now i think this will work for people who want to load britecharts NOW.

## Motivation and Context
cdns are cool!

## How Has This Been Tested?
cdn.html in demos

image


dist/css/bundle.css
```diff
@@ -0,0 +1,147 @@
+@font-face {
```
ryanwholey: I'm not sure where this css bundle came from actually, is this old? It didn't work correctly with our charts, went with britecharts.css instead. I'm going to remove it if thats cool.......

webpack.config.js
```diff
@@ -165,6 +165,39 @@ config = {
...
+            // Tell Webpack not to parse certain modules.
+            noParse: [
+                new RegExp(vendorsPath + '/underscore/underscore.js'),
+                new RegExp(vendorsPath + '/d3/d3.js')
+            ]
+        },
+
+        resolve: {
+            alias: {
+                underscore: vendorsPath + '/underscore/underscore.js',
```
ryanwholey: do we need this? shouldnt webpack know to find underscore in our node_modules folder already? i'm more familiar with aliasing common app paths

ArkaAblion:
hi,I am trying to show the tooltip on barchart but could not make it done,please help.
I am using Jquery,CDN LINKS for britecharts .Here is my code......
<script type="text/javascript">
             // Bar
            var barChart = britecharts.bar();
            var tooltip = britecharts.tooltip();
            var barData = [
                            {name: 'Jan', value:150000},
                            {name: 'Feb', value: 45000},
                            {name: 'Mar', value: 85000},
                            {name: 'Apr', value: 190000},
                            {name: 'May', value: 200000},
                            {name: 'June', value: 220000},
                            {name: 'July', value: 85000},
                            {name: 'Aug', value: 170000},
                            {name: 'Sep', value: 160000},
                            {name: 'Oct', value: 220000},
                            {name: 'Nov', value: 300000},
                            {name: 'Dec', value: 200000}
                        ];
            barChart
                 .margin({left: 100,right: 5,top: -10,bottom: 20})
//                .colorSchema(britecharts.colors.colorSchemas.britechartsColorSchema)
                .isAnimated(true)
                .percentageAxisToMaxRatio(1.3)
                .width(760)
                .height(500)
                .on('customMouseOver', tooltip.show)
                .on('customMouseMove', tooltip.update)
                .on('customMouseOut', tooltip.hide);
        tooltip
                .title('CDN tooltip');
                d3.select('.bar_chart').datum(barData).call(barChart);
                d3.select('.bar_chart').datum(barData).call(tooltip);
       </script>

PR #245, opened 2017-05-18
Feature highlight biggest slice

tgallice:
## Description
By default the donut doesn't have any slice `highlighted`. This feature add the possibility to highlight by default the biggest slice. Like that there is always a value displayed in the donut.

I'm not very happy with the naming, if you have some idea feel free to tell me.

## Motivation and Context
Personnal need :) 

Todo:
- [ ] Generate demos && documentation
- [ ] Generate UMD

Default with the mouse out of the donut:
capture du 2017-05-18 21-20-52

When the mouse is over an other slice:
capture du 2017-05-18 21-21-11

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -85,6 +85,10 @@ define(function(require){
             slices,
             svg,
 
+            // highlight slice
+            highlightedSliceId,
+            highlightedSlice,
+
             quantityLabel = 'quantity',
             nameLabel = 'name',
             percentageLabel = 'percentage',
@@ -128,6 +132,7 @@ define(function(require){
                 buildSVG(this);
                 drawSlices();
                 initTooltip();
+                initHighlightSlice();
             });
         }
 
@@ -251,8 +256,8 @@ define(function(require){
                 .merge(slices)
                   .append('path')
                     .attr('fill', getSliceFill)
-                    .on('mouseover', tweenGrowthFactory(externalRadius, 0))
-                    .on('mouseout', tweenGrowthFactory(externalRadius - radiusHoverOffset, pieHoverTransitionDuration))
+                    .on('mouseover', handlePathMouseOver)
+                    .on('mouseout', handlePathMouseOut)
                     .transition()
                     .ease(ease)
                     .duration(pieDrawingTransitionDuration)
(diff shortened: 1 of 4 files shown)


Golodhros:
Hi @tgallice , thanks a lot for your work in this PR, it is great work!

I understand your requirement and we thought about this feature before. The implementation is also great and we would love to have something like this, but we would love this to be more generic.

By generic I mean that your solution works great for the use case that we want to highlight the largest slice, but it won't work if we wanted to highlight the smallest slice or another one in the middle. 

That's why I wanted to ask you to modify your solution a bit. We think that a more generic approach would use the 'id' of the entry as an identifier. The final API would be something like:
  
  donut.highlightSliceById(2);

This way, we could choose any slice to highlight and would avoid to create other selection methods like 'highlightSmallestSlice'. 

The downside of it is that we would force our users to process a bit more their data. In the case of selecting the largest slice, this would mean ordering the data by size, add ids and use `highlightSliceById` to highlight the proper slice.

What do you think about it? Would that work for you?

tgallice:
@Golodhros You are absolutely right, it's a better solution! I have made modifications according to your comment.

Let's me know if I have missing or need to rework somes parts.

Thank for your feedback.

src/charts/donut.js
```diff
@@ -282,18 +287,60 @@ define(function(require){
             svg.select('.donut-text').text('');
         }
 
+        /**
+         * Find the slice by id and growth it if needed
+         * @private
+         */
+        function initHighlightSlice() {
+            if (!highlightedSliceId && highlightedSliceId !== 0) return;
```
Golodhros: I was thinking that maybe we could compare `highlightedSliceId` with undefined, so that indexes from 0 to N would be valid.

Nit: Could you add curly braces? We usually don't do one-lines.
Thanks!

src/charts/donut.js
```diff
@@ -282,18 +287,60 @@ define(function(require){
...
+
+            highlightedSlice = svg.selectAll('.chart-group .arc path')
+                .select(function(a) {
+                    if (a.data.id === highlightedSliceId) {
+                        return this;
+                    }
+                }).node();
+
+            if (!highlightedSlice) return;
```
Golodhros: Nitpick: We would prefer
  if (highlightedSlice) {
     ...code
  }

thanks!

Golodhros:
Hi @tgallice , thanks a lot for your work! I have been checking the code and looks great!

I was testing in the demo and I was wondering, your requirements include to keep the highlighted slice on?  

At first, I was thinking about just highlighting it by default, but once we hover on another slice, I was expecting it to proceed as usual. If you need it to be locked on the highlighted slice, maybe we could separate that behavior, maybe in a 'fixHighlightedSlice' configuration.

Let me hear your thoughts. Thanks a lot!

tgallice:
Hey @Golodhros sorry for the delay, I'm a bit busy :)

I'm not sure to understand you, the highlighted slice is "highlighting" only when the cursor is `out` the donut or `on` the highlighted slice, no ? If the cursor is on an other slice, the highlighted slice should be in the "flat" state. There is something wrong ?

Golodhros:
Hi @tgallice,
Well, is not really a problem, just a unexpected result.
We were thinking on highlighting the slice only when we first load the chart. After any interaction, we want it to behave as normal, and not highlight the slice again.

If you really need to have that interaction, I would suggest adding another property, 'hasFixedHighlightedSlice', so that we lock the highlighted slice on that state.

Are you OK with this?

Golodhros:
Merged in https://github.com/eventbrite/britecharts/pull/271

Thanks for your work!!

PR #246, opened 2017-05-18
Skip data with zero quantity

tgallice:
## Description
Skip data without quantity.

## Motivation and Context
If we have a data without quantity, this should be skip otherwise the donut will display an `empty` slice but the margin will lead a space (cf screenshot). A pre-processing can by apply on the data before using it, but this will break the portability of this dataset in the `legend` chart.

TODO:
- [ ] Update docs
- [ ] Update UMD

Donut with data without quantity.

capture du 2017-05-18 22-18-45

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -205,13 +205,19 @@ define(function(require){
          * @private
          */
         function cleanData(data) {
-            return data.map((d) => {
+            return data.reduce((acc, d) => {
+                // Skip data without quantity
+                if (!d[quantityLabel]) {
+                    return acc;
+                }
+
                 d.quantity = +d[quantityLabel];
                 d.name = String(d[nameLabel]);
                 d.percentage = String(d[percentageLabel]);
+                acc.push(d);
 
-                return d;
-            });
+                return acc;
+            }, []);
         }
 
         /**


Golodhros:
Hi @tgallice , thanks a lot for your work in this PR and previous ones! 

In theory, the quantity is a required value for each entry for the donut chart. Maybe I am missing something, but I don't know how would be a use case where one of the entries don't have any quantity.

Could you tell us a bit more about it?

Thanks a lot!

tgallice:
If my dataset is something like that:
```js
let dataset = [
  { name: 'Data 1', quantity: 1, percentage: 25, id: 1 },
  { name: 'Data 2', quantity: 3, percentage: 75, id: 2 },
  { name: 'Data 3', quantity: 0, percentage: 0, id: 3 },
];
```
I need to apply a filter to exclude the `Data 3` in the donut chart (otherwise a space will appear like the screenshot), but need to keep it for the legend one. To me, this dataset should be valid for the donut chart without break the rendering, no ?

This fix is probably not the best one. There is a way to not add `padding` to slice with 0 quantity ? I'm not a D3 expert :/

miglesiasEB:
Hi Thomas,

I have been discussing with @evb-sun about this issue. Maybe the right solution would be to actually remove the padding between the donut slices. 

We think that would work your case, and also for the case where there is only one group, as we are showing it as one slice that has a small white line (the margin).

What do you think?

Golodhros:
Merged in #246 

Thanks!

PR #253, opened 2017-05-21
Fix docs merged

bung87:
## Description

merged my prs,and fix my grunt task config problems,the end results is 
  1. use webpack-dev-server serves on ./docs path,let "jsdoc" layout template include /assets/bundle.js and config webpack publicPath to '/assets/',at the end when run command `npm run demos:serve` will serve docs at localhost and open it via default browser without "/docs" path.
  2. the docs index.html which use readme as a doc section contains "api","demo" links will refferences to localhost and no side effect to readme on github.  

there's still some problems leaves here:
<del>1. the demo page contains "data input schema" link still refferences to github.io page may change it use template variable use a  different jsdoc config,or manipulate in publish.js.  </del>
2. since the webpack-dev-server hot reload bundle.js ,other resources related to docs do not have hot reload feature,may use other tool to startup a socket or reuse the webpack-dev-server's socket.  

## Types of changes

## Checklist:


src/tasks/js-doc.js
```diff
@@ -17,35 +17,33 @@ module.exports = function(grunt) {
...
+    // });
 
     grunt.registerTask('docs', 'Generates docs and triggers server to view them', [
         'jsdoc:dist',
-        'moveDemos',
-        'moveFonts',
-        'openDocs'
+
+        'sync'
+        // 'moveDemos',
```
Golodhros: I understand that we do this now with Grunt sync, right?

src/doc/template/tmpl/layout.tmpl
```diff
@@ -161,13 +161,13 @@
             </span>
         </footer>
 
-        <script src="scripts/docstrap.lib.js"></script>
-        <script src="scripts/toc.js"></script>
+        <script src="/britecharts/scripts/docstrap.lib.js"></script>
```
Golodhros: I don't think that this will work once we push it into our github pages, as the parth would be `/docs/scripts`

src/tasks/js-doc.js
bung87: yes,it's grunt-sync.

src/doc/template/tmpl/layout.tmpl
bung87: github.io serves docs/ directory as http://eventbrite.github.io/britecharts/ so it would not be a problem.

src/tasks/js-doc.js
Golodhros: Thanks!

src/doc/template/tmpl/layout.tmpl
Golodhros: Great!

Issue #243, opened 2017-05-17
Option to disable animations


AlfonsoML:
At the moment the way to make the charts responsive is to redraw them. That happens also when there's some change.  
The problem is that each time the whole chart gets a "loading" animation that draws it and it might be useless after the initial creation.

## Expected Behavior
Have a way to redraw the chart without extra distractions, show the chart ASAP.

## Current Behavior
Each time that the chart is created it always use a "loading" animation starting from 0 to max

Golodhros:
HI @AlfonsoML , thanks for the issue!

That's a totally legit request, not the first time I think about it!
It should be pretty straightforward to add an accessor to stop the animation, so that when we redraw the chart, we set to not animate and get the desired outcome.

Thanks a lot!

Golodhros:
Fixed on #257

PR #257, opened 2017-05-22
Adding animations configuration to Donut, Line and Stacked Area Charts

Golodhros:
Adding animations configuration to Donut, Line and Stacked Area Charts

## Description
We wanted to add the option of switch on/off the animations on the charts. In this PR, we adress the charts without them (sparkline had it already).

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/243

## How Has This Been Tested?
Ran tests

## Types of changes

## Checklist:


src/charts/donut.js
```diff
@@ -241,23 +242,38 @@ define(function(require){
...
-                    .attr('fill', getSliceFill)
-                    .on('mouseover', tweenGrowthFactory(externalRadius, 0))
-                    .on('mouseout', tweenGrowthFactory(externalRadius - radiusHoverOffset, pieHoverTransitionDuration))
-                    .transition()
-                    .ease(ease)
-                    .duration(pieDrawingTransitionDuration)
-                    .attrTween('d', tweenLoading);
-
+                if (isAnimated) {
+                    slices.enter()
```
rwholey-eb: what if we stored the value of these outside of the if block? DRY stuff out?

src/charts/stacked-area.js
```diff
@@ -536,35 +537,57 @@ define(function(require){
...
-                .data(layers)
-                .transition()
-                .delay( (_, i) => areaAnimationDelays[i])
-                .duration(areaAnimationDuration)
-                .ease(ease)
-                .attr('d', area)
-                .style('opacity', areaOpacity)
-                .style('fill', ({key}) => categoryColorMap[key]);
+            if (isAnimated) {
+                series = svg.select('.chart-group').selectAll('.layer')
```
rwholey-eb: same here

src/charts/donut.js
Golodhros: You mean the first block, before the merge?

Let me try, but I think it won't work...

src/charts/stacked-area.js
Golodhros: In this case they are slightly different (layersInitial vs layers), so I cannot extract it.

rwholey-eb: got it. thanks!

PR #262, opened 2017-05-23
Update indents in Donut.js

hckrmoon:
## Description

I am not sure about this fixture about indents.
Is it you guys's convention?

### DIFF
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -148,7 +148,7 @@ define(function(require){
          */
         function buildContainerGroups() {
             let container = svg
-              .append('g')
+                .append('g')
                 .classed('container-group', true)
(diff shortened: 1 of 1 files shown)


nobitagit:
I actually like the fact that Britecharts uses that convention (see [here](https://bost.ocks.org/mike/d3/workshop/#35) and [here](https://github.com/alignedleft/d3-book/issues/1)) , it makes it easier to see which selection will be affected by the code that follows.

Golodhros:
Hi @hckrmoon, 
As @nobitagit correctly pointed out, this is a common convention in the D3 world. The idea is to 'mark' through the indentation when a seletion context has changed.

I will close this issue now. Thanks for contributing anyways!

hckrmoon:
Thank you for your explanation ! :D

PR #266, opened 2017-05-24
Bar Chart Animations with flag

Golodhros:
Adding the missing animations on bar charts, both horizontal and vertical

## Description
I am adding animations to the bar charts and improving the demo, as well as adjusting the line chart demo spacing and load to fit better on mobile sizes.

## Motivation and Context
Related to #243, I think we should add animations on the bar chart for the sake of completion.

## How Has This Been Tested?
Ran tests and checked on demos

## Types of changes

## Checklist:


demos/demo-line.js
```diff
@@ -10,12 +10,14 @@ var _ = require('underscore'),
     line = require('./../src/charts/line'),
     tooltip = require('./../src/charts/tooltip'),
     dataBuilder = require('./../test/fixtures/lineChartDataBuilder'),
-    colorSelectorHelper = require('./helpers/colorSelector');
+    colorSelectorHelper = require('./helpers/colorSelector'),
+
+    barMargin = {top:60, bottom: 50, left: 50, right: 30};
```
rwholey-eb: barMargin, this is line demo?

src/charts/bar.js
```diff
@@ -317,6 +328,81 @@ define(function(require) {
         }
 
         /**
+         * Draws and animates the bars along the x axis
+         * @param  {D3Selection} bars Selection of bars
+         * @return {void}
+         */
+        function drawAnimatedHorizontalBars(bars) {
+            // Enter + Update
```
rwholey-eb: what about trying to combine some of the logic here? I see this divergent pattern as a potential problem in the future, where horizontal and vertical end up with different behaviors/features

demos/demo-line.js
Golodhros: Oops!! Good call!

src/charts/bar.js
Golodhros: Yeah, totally. I am not super happy with this approach either.
I will think about how to do it other way...

PR #302, opened 2017-07-08
Normalizing chart names between umd and bundle versions. Fixes #288

### LINKED ISSUE #288: Stacked-bar from dist/bundled/britecharts and from dist/umd doesn't work the same
Hello,

The stacked-bar chart from dist/bundled/britecharts.min.js and the stacked-bar from dist/umd/stacked-bar.min.js doesn't work the same

## Expected Behavior
In all case, the stacked-bar constructor should be stackedBar(), for now it's `stacked-bar()` in umd file
`chartContainer.datum(dataset).call(barChart); ` shouldn't trigger an error

_Exemple with bundled file_
https://codepen.io/sound-matt/pen/weJdMJ

_Exemple with UMD files_
https://codepen.io/sound-matt/pen/gRPmXK

Golodhros:
## Description
Changing the names on the chart list we use to create the UMD versions of our charts.

## Motivation and Context
Chart names weren't the same on the UMD and bundle versions as described in #288 

## How Has This Been Tested?
Ran tests and bundled production.

## Types of changes

## Checklist:


Golodhros:
@ryanwholey This stuff could be annoying for some of our users, but makes sense in the long term.

rwholey-eb:
When do you think would be a good time to land this?

Golodhros:
Any time now, unless you needed the previous fix for the reports!

mmartimo:
Just wanted to check in on the release that includes this...   which as I understand opens up full support via the CDN?   I'm in need of grouped bar charts, pie chart legends, etc...  unless there is a way to access those features via CDN now? (acknowledging that it will need to be updated for the next release)

Golodhros:
Hi @mmartimo,
It could be that the next release takes a bit of time, as we are looking into API changes --explained in https://github.com/eventbrite/britecharts/issues/311.

At this moment, you can check what is included on the CDN repo by following this link: https://cdn.jsdelivr.net/npm/britecharts@1.7.2/dist/umd/

Issue #264, opened 2017-05-23
Create a code styleguide for Britecharts


Golodhros:
We want to create a code styleguide for Britecharts that could make more straightforward code reviews, discussions and questions. It would be also helpful for new contributors that don't feel confortable about the way they are coding their solutions.

## Expected Behavior
We should have a code styleguide to reference when talking about code on PRs, issues and code reviews.

## Current Behavior
We do not have that.

## Possible Solution
Create a markdown file with some of those, including:
* Indentation 
* Margin convention
* Testing patterns
* Ternary vs multiple if-else
* General best practices

## Context
We want to make Britecharts a project that is easy to contribute to.

gagoar:
It might be great if we could present some sort of structure for each coding style. 

I think the one we use for our coding style could work. its basic structure is:

- title (the term you would most likely call the code style, like *indentation*)

- motivation (reasons why this style exist)

- the good (how should look like)

- the bad (how should not look like)

- case study (if you see that might be a good idea to show the reasoning for it)

take a look at [this example](https://github.com/eventbrite/javascript/tree/master/general#ternary-statements)

nobitagit:
Some of those points can be addressed by wiring up the `eslintrc.json` file to the build so that any PR that does not conform to the rules automatically fails as soon as it is opened.

It won't be possible to address all of the issues in that sense but it will certainly cut down the efforts on your side while doing code reviews.

Also, I noticed that the Contributing.md file could be in need of some updating from you guys. It mentions tasks like `npm run dev`that are not defined in the package, so perhaps this could be tackled as well.

gagoar:
correct! whatever we can address with `eslint` should be done in that fashion, but I think here, there are several things that are more stylistic and not always easy to spot with `eslint`. (altho arguably you could do almost anything with eslint).

Golodhros:
Related to this, I just created a PR to clean our eslint issues: https://github.com/eventbrite/britecharts/pull/312

PR #312, opened 2017-07-23
Updating eslint config and cleaning issues

Golodhros:
## Description
Our code wasn't in sync with our eslint rules. This PR updates that eslint configuration and fixes all the issues. It also updates the CodeClimate configuration.

## Motivation and Context
This PR will allow us to run the linter with our builds and it will probably fix the linting issues showing on CodeClimate.
Also, related to https://github.com/eventbrite/britecharts/issues/264

## How Has This Been Tested?
Ran eslint in the ```src/charts``` folder

## Types of changes

## Checklist:


src/charts/step.js
```diff
@@ -370,16 +372,30 @@ define(function(require) {
...
-         * Gets or Sets the number of vertical ticks on the chart
+         * Exposes an 'on' method that acts as a bridge with the event dispatcher
+         * We are going to expose this events:
+         * customMouseOver, customMouseMove and customMouseOut
+         *
+         * @return {module} Bar Chart
+         * @public
+         */
+        exports.on = function() {
+            let value = dispatcher.on.apply(dispatcher, arguments);
```
entobot: In modern JavaScript, the `arguments` object is considered unnecessary and bad practice, since you can achieve the same behaviour by using the more easily understandable [spread operator](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator). 

You can rewrite the code above as follows:

```javascript
        exports.on = function(...args) {
            let value = dispatcher.on(...args);
```

src/charts/bar.js
```diff
@@ -100,13 +100,14 @@ define(function(require) {
             isAnimated = false,
             ease = d3Ease.easeQuadInOut,
             animationDuration = 800,
-            interBarDelay = function(d, i) {return 70 * i},
+            animationStepRatio = 70,
+            interBarDelay = (d, i) => {animationStepRatio * i},
```
entobot: This function has no side effects and does not return a value. In effect, it does nothing.

As the function consists of a single expression, I am guessing it's supposed to return the result of that expression. If this is the case, I would suggest making this change:

```javascript
            interBarDelay = (d, i) => animationStepRatio * i,
```

src/charts/brush.js
```diff
@@ -400,7 +400,7 @@ define(function(require) {
          * @return {void}
          */
         function updateHandlers(dateExtent) {
-            if (dateExtent == null) {
+            if (dateExtent === null) {
```
entobot: Comparing against `null` with the `===` operator only checks that the value is `null`, but not `undefined`. If you would like to check for both `undefined` and `null`, you can simply use `==`. This will coerce them to the same value, which means that you just need to check against `null` for both.

Nothing else will ever be coerced to `null` or `undefined`, so this is the one occasion where `==` is safe to use.

If this is the case, I would suggest making this change:

```javascript
            if (dateExtent == null) {
```

src/charts/helpers/text.js
```diff
@@ -24,7 +25,7 @@ define(function(require) {
      * More discussions on https://github.com/mbostock/d3/issues/1642
      * @return {void}
      */
-    const wrapText = function(xOffset, fontSize, availableWidth, node, data, i) {
+    const wrapText = function(xOffset, fontSize, availableWidth, node) {
```
Golodhros: It's fine this way, thanks!

src/charts/stacked-bar.js
```diff
@@ -289,9 +289,9 @@ define(function(require){
 
             categoryColorMap = colorScale
                 .domain(data.map(getName)).domain()
-                .reduce((memo, item, i) => {
+                .reduce((memo, item) => {
                     data.forEach(function(v){
-                        if (getName(v)==item){
+                        if (getName(v) == item){
```
Golodhros: Makes sense, thanks!

src/charts/stacked-bar.js
```diff
@@ -433,14 +433,13 @@ define(function(require){
...
+                            i = d3Interpolate.interpolateRound(0,xScale(d[1] - d[0])),
+                            j = d3Interpolate.interpolateNumber(0,1);
 
-                        return function(t){
+                        return function(t) {
                             node.attr('width',  i(t) );
                             node.style('opacity', j(t) );
-                        }
+                        };
                     });
```
Golodhros: Good call, still it's not big enough to extract into a different function.

src/charts/step.js
Golodhros: Good call!

src/charts/bar.js
Golodhros: Good call, this error wasn't catch by our tests.

src/charts/brush.js
Golodhros: Thanks, this is good for now.

PR #334, opened 2017-08-14
Fix color update

biovisualize:
When updating the color using `colorSchema`, the getter was giving the right value, but the DOM was not updated with the new color palette. 

## Description
I just moved the `fill` attribute from `enter`to after `merge`

## How Has This Been Tested?
I tested on a small local fixture, using a timer to change the color after a second. I also made a unit test.

## Types of changes

## Checklist:


test/specs/bar.spec.js
```diff
@@ -72,6 +72,20 @@ define(['d3', 'bar', 'barChartDataBuilder'], function(d3, chart, dataBuilder) {
...
+                let previous = barChart.colorSchema(),
+                    expected = '#FFFFFF',
+                    actual;
+
+                barChart.colorSchema([expected]);
+
+                const barColor = containerFixture.select('rect.bar');
+
+                containerFixture.call(barChart);
+
```
Golodhros: Nit: We would do here:
  actual = barColor.attr('fill');

  expect(actual).toBe(expected);

Golodhros:
Thanks for the PR @biovisualize!

PR #339, opened 2017-08-27
Add lineCurve public attribute to Line Chart

dylanmoz:
## Description
Add the ability for line charts to have curves other than linear. This PR exposes a public method `lineCurve`, which takes a d3 curve factory (such as `d3Shape.curveBasis`). The default value is `d3Shape.curveLinear`. The biggest difficulty with this change was correctly positioning the `data-point-highlighter` circle for each line on mouse over. The problem is that you can no longer position the circle at `yScale(value)`, since a line curve may not be at that point anymore, so the circle would awkwardly be dangling away from the line. To solve this, I had to traverse each path using bisection to find the appropriate y value given a dataPoint date.

## Motivation and Context
Allows for more customization of line charts, when users don't want their line curves to have sharp edges, for example.

## How Has This Been Tested?
Wrote test for `lineCurve` existence.
Test with various curves.
Tested performance of `highlightDataPoints` before and after changes. Before was ~2.68ms, after ~6.07ms. Added `pathYCache` to reduce number of computations, reducing the time to ~3.15ms.

screen shot 2017-08-27 at 10 23 53 am

screen shot 2017-08-27 at 1 24 39 pm

## Types of changes

## Checklist:


src/charts/line.js
```diff
@@ -721,11 +728,24 @@ define(function(require){
         function highlightDataPoints(dataPoint) {
             cleanDataPointHighlights();
 
+            const nodes = paths.nodes()
+            const nodesById = nodes.reduce((acc, node) => {
+                acc[node.id] = node
+                return acc
```
Golodhros: Nit: let's add semicolons and a space before return

src/charts/line.js
```diff
@@ -721,11 +728,24 @@ define(function(require){
...
-                                    .sort((a, b) => topicColorMap[a.name] < topicColorMap[b.name]);
+            const topicsWithNode = dataPoint.topics
+                                        .map(topic => ({
+                                            topic,
+                                            node: nodesById[topic.name]
+                                        }))
+                                        .filter(({topic}) => !!topic)
+                                        .sort((a, b) => topicColorMap[a.topic.name] < topicColorMap[b.topic.name])
+
+            dataPoint.topics = topicsWithNode.map(({topic}) => topic)
```
Golodhros: Nit: semicolons

src/charts/line.js
```diff
@@ -740,11 +760,53 @@ define(function(require){
                     .attr('r', 5)
                     .style('stroke', topicColorMap[name]);
 
-                marker.attr('transform', `translate( ${(- circleSize)}, ${(yScale(dataPoint.topics[index].value))} )` );
+                const path = topicsWithNode[index].node;
+                const x = xScale(new Date(dataPoint.topics[index].date));
+                const key = `${name}-${x}`;
+                const y = key in pathYCache ? pathYCache[key] : (pathYCache[key] = findLineYatX(x, path));
```
Golodhros: Nit: This line got a bit too complex. Could you use a function?

src/charts/line.js
```diff
@@ -740,11 +760,53 @@ define(function(require){
...
+
+                point = path.getPointAtLength(midpoint);
+
+                if (x < point.x) {
+                    lengthEnd = midpoint;
+                } else {
+                    lengthStart = midpoint;
+                }
+
+                iterations += 1
```
Golodhros: semicolon

test/specs/line.spec.js
```diff
@@ -409,6 +409,18 @@ define([
                     expect(actual).toBe(expected);
                 });
 
+                it('should provide lineCurve getter and setter', () => {
+                    let previous = lineChart.lineCurve(),
+                        expected = d3.curveBasis,
```
Golodhros: Oops, I think this needs an update too.
Thanks

dylanmoz: Ah. Good catch

Golodhros:
Awesome work @dylanmoz !! 
Thanks for your contribution!

Issue #291, opened 2017-06-22
Refreshing Graph Data


rpheath:
I render a brush chart initially via a callback from an XHR request, which works perfectly. However, in that callback, I set a timer to "refresh" the chart in N seconds, to keep the chart/data fresh. The first render goes smoothly, but subsequent renders (via the timer) seems to 'stack' the data instead of reset/refresh it.

## Expected Behavior
I would not have expected the data to stack up, but instead when I call `container.datum(data).call(@chart)` the data for the chart would have been reset to whatever the `data` collection contained.

## Current Behavior
The data stacked up so I was showing multiple data sets in the same graph 🤔.

## Possible Solution
Make it so the data handed to `@container.datum(data)` is the exclusive data set for the graph.

## Steps to Reproduce (for bugs)
Example pseudo code:

```coffeescript
loadGraph: ->
  $.get '/my/path.json', { ... }, @graphHandler

graphHandler: (data) =>
  container = d3selection.select('#my-container')
  container.datum(data).call(@chart) # <-- is this not the proper way to set/refresh data?

  _.delay =>
    @loadGraph()
  , 5000
```

## Context
I need to be able to easily refresh a chart with new data.

## Environment
* Version used: 1.5.3
* Browser Name and version: Safari (latest), Chrome (latest)
* Operating System (desktop): macOS Sierra

rpheath:
Ahhh, as it always goes, I've realized my problem after posting this issue!

I'm going to dump the explanation here in case someone else stumbles into this:

I misunderstood the API in that the you have to instantiate a new chart on **EVERY** data refresh. That seemed unintuitive to me, since the chart itself seems to represent the properties of the chart and defines the callbacks (if they exist) on interactions, but the data (I thought) was independent of the chart itself. That was my mistake.

I was instantiating the chart ONE time in the constructor of my class, and then trying to `@container.datum(data).call(@chart)` where `@chart` was that same chart that was already instantiated. This is what was causing the data to stack up.

After realizing that I had to set the data on a `new BrushChart()` each time, it seems to work as I need it to. It may be worth mentioning that in the docs (or being explicit about that in the code examples), as data-refreshing is a common need in charting/graphing.

Closing this issue and apologies if this wasted anyone's time.

rpheath:
[...]

PR #351, opened 2017-09-11
Fixing data reloads on Donut Chart

Golodhros:
## Description

Fixing some code weirdness on the Donut implementation.
Fixing eslint warnings.
Fixing js sourcemaps thanks to @jchen-eb 

## Motivation and Context

Reloading data on the same donut chart instance wasn't working properly, resulting on an empty chart once it got reload. This is part of https://github.com/eventbrite/britecharts/issues/291.

## How Has This Been Tested?

Added new tests and checked manually by modifying the demo and loading new data.

## Types of changes


src/charts/donut.js
```diff
@@ -225,9 +229,8 @@ define(function(require) {
                 d.quantity = +d[quantityLabel];
```
rwholey-eb: could we use object spread here too!? I remember running into some build issues when i tried a while back, maybe its fixed now

src/charts/donut.js
```diff
@@ -269,63 +272,54 @@ define(function(require) {
...
+                .append('g')
+                  .each(storeAngle)
+                  .each(reduceOuterRadius)
+                  .classed('arc', true)
+                  .append('path');
+        
+            if (isAnimated) {
+                newSlices.merge(slices)
+                    .attr('fill', getSliceFill)
+                    .on('mouseover', function(d) {
```
rwholey-eb: could we use the functions by their name instead of wrapping them in function calls? We could bind if this is an issue

webpack.config.js
```diff
@@ -92,7 +92,7 @@ config = {
 
     // Add here listener to sccs files?
     demos : {
-        devtool: 'source-map',
+        devtool: 'cheap-eval-source-map',
```
rwholey-eb: faster, nice. I found eval to give me more accurate stack traces but i haven't investigated enough to know exactly why. If this doesn't do what we want, may be worth investigating

src/charts/donut.js
Golodhros: I will give it a try

webpack.config.js
Golodhros: This works nicely now, I did't know about it, it was @jchen-eb

src/charts/donut.js
Golodhros: I couldn't figure it out!
The problem is that if I bind, I lose the 'this' inside this callback, that is the specific donut slice that was hovered over. If I try to pass it down it comes as 'undefined'.
🤷‍♂️

PR #356, opened 2017-09-17
Fixing data reload on Grouped Bar Chart

Golodhros:
## Description

Part of the fixes for https://github.com/eventbrite/britecharts/issues/291 
I am removing the current elements when there is a reload of data. This solution is not ideal, we should be using .exit(), but I couldn't make it work as there are nested elements. This also happened on the donut chart, so both fixes are temporal.

## How Has This Been Tested?

Added tests

## Types of changes


docs/cdn-iframe.html
```diff
@@ -8,7 +8,7 @@
 
         <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/4.4.4/d3.min.js"></script>
         <!-- CDN is still a WIP, when it's ready the js cdn bundle should be inserted beow! -->
-        <script src="https://cdn.jsdelivr.net/npm/britecharts@2/dist/bundled/britecharts.min.js" type="text/javascript"></script>
+        <script src="https://cdn.jsdelivr.net/npm/britecharts@1/dist/bundled/britecharts.min.js" type="text/javascript"></script>
```
rwholey-eb: whats going on with our path here? @1 vs @2?

Golodhros: Oh, this is wrong! Good catch!

Golodhros:
More or less!
With this change, we will be able to do:
  container.data(dataSet).call(stackedAreaChart);

and later do:
  container.data(dataSet2).call(stackedAreaChart);

with different data and it should work fine, removing the old entries without creating a new chart.

Issue #305, opened 2017-07-11
Add Dots to Stacked Area Chart with dot click Interactivity


rpheath:
Is it possible to add custom interactions to the charts at all? Specifically, and as an example, let's use a stacked area chart... I'm looking to provide options based on specific points on the graph. Let's say I wanted to bind a click event to each "dot" that shows up on hover, and when you clicked it, you got a menu of options to take action based on that data.

Is there a recommended path to accomplish that?

## Possible Solution
I could see giving a handler to the chart the same way the `.on('customerMouseOver', tooltip.show)` stuff works. Could be something like:

```coffeescript
chart
  .on('customClick', handlers.click)
```

(Does anything like that exist?)

## Context
Being able to interact (outside of a brush chart), even if very basically, with different types of charts (bar, area, etc) is going to be a big factor into whether or not we can use Britecharts in our project.

Thanks!

Golodhros:
Hi @rpheath, thanks for the issue!

That's an interesting use case. That could happen but is not done right now. It would be something like showing the dots and adding an event handler to those dots. That could be mapped on to a different custom event as you suggest. I would probably call it `customDataEntryClick` or similar.

Which charts would you need to have this feature? I think it makes sense for the Stacked Area chart. Maybe on the LineChart too.

rpheath:
I like `customDataEntryClick` more, too. And yes, the Stacked Area and Line chart's would be the two that come to mind 👍

rpheath:
Additionally, though, I could see value in being able to "click a bar" in a bar chart, as well. But my specific use case currently are with the area/line chart dots.

Golodhros:
@rpheath makes sense!

I will try to find some time to do this. Sadly, the bar chart case is much easier than the other one, but it makes sense to add this functionality.

Thanks,

rpheath:
@Golodhros glad to see this added to the backlog! Now, any idea how far away 3.0 is?! 😉

rpheath:
Also, I wanted to note that as part of the implementation (at least for my use case), being able to access the "data" of the clicked dot is crucial, so if that could be handed off to the click event as parameters (or at least a recommended path to achieve that), that would be suhhweeeet!

Golodhros:
Hi @rpheath,

So it's hard to say when the 3.0 will be ready, but you will be able to have this feature before that. We are focusing now on fixing bugs and the React wrapper, but the initial simplicity of this feature means that we could we tackle it soon.

[...]

PR #364, opened 2017-09-26
Adding dots to Stacked Area and Line Chart (unfinished)

Golodhros:
## Description

Adding click events to the Stacked Area and Line Chart highlight circles. 
It's working fine in the Stacked Area, but I can't spot the problem on the Line Chart.

## Motivation and Context

Related to https://github.com/eventbrite/britecharts/issues/305. We are adding the ability for our users to trigger actions when the entry points are being clicked/touched.

## How Has This Been Tested?

I couldn't figure out the way to test this, as we should be able to first trigger a mousemove and then a mouse click within the exposed circles.

## Types of changes


Golodhros:
Implemented in https://github.com/eventbrite/britecharts/pull/385

PR #371, opened 2017-10-02
Fixing data reload on stacked bar chart

Golodhros:
## Description

We are cleaning the old layers before redrawing the new data. This is still the non-optimal solution, but it will work for now.

## Motivation and Context

Related to https://github.com/eventbrite/britecharts/issues/291, this is part of the fixes for that issue.

## How Has This Been Tested?

Added tests

## Types of changes


src/charts/stacked-bar.js
```diff
@@ -411,46 +413,42 @@ define(function(require){
 
         /**
          * Draws the bars along the x axis
-         * @param  {D3Selection} bars Selection of bars
+         * @param  {D3Selection} layersSelection Selection of bars
          * @return {void}
          */
-        function drawHorizontalBars(series) {
+        function drawHorizontalBars(layersSelection) {
+            let layerJoin = layersSelection
```
rwholey-eb: new repo! any interest in switching to const when possible?

src/charts/stacked-bar.js
```diff
@@ -473,44 +471,42 @@ define(function(require){
...
+        function drawVerticalBars(layersSelection) {
+            let layerJoin = layersSelection
+                .data(layers);
+
+            layerElements = layerJoin
                 .enter()
                   .append('g')
-                    .classed('layer', true)
                     .attr('fill', (({key}) => categoryColorMap[key]))
+                    .classed('layer', true);
```
rwholey-eb: layer seems generic, could this have conflicts with other charts down the road?

src/charts/stacked-bar.js
```diff
@@ -473,44 +471,42 @@ define(function(require){
...
+            // Enter + Update
+            let bars = barJoin
                     .enter()
                       .append('rect')
                         .classed('bar', true)
                         .attr('x', (d) => xScale(d.data.key))
                         .attr('y', (d) => yScale(d[1]))
                         .attr('width', xScale.bandwidth )
-                        .attr('fill', (({data}) => categoryColorMap[data.stack+data.key])),context;
+                        .attr('fill', (({data}) => categoryColorMap[data.stack+data.key]));
```
rwholey-eb: could we do string interpolation here? `${data.stack}${data.key}`, for a second i thought you might be adding numbers

src/charts/stacked-bar.js
Golodhros: It's not on my list of priorities, although I know Jaylum is more used to use const.

src/charts/stacked-bar.js
Golodhros: Mmmm... good call. That could happen yes, I will keep an eye in this.

src/charts/stacked-bar.js
Golodhros: Totally!

jchen-eb:
LGTM, nice fix!

PR #381, opened 2017-10-10
Adding topicsOrder option to stacked area chart

Golodhros:
## Description

We are allowing to set the order of the tooltip elements, but until now we weren't able to change the stacking order of the stacked area chart layers. This commit fixes that.

## Motivation and Context

Sometimes we want to force a stacking order that makes more sense to the users. It also allows for feature parity with the tooltip.

## How Has This Been Tested?

Added a simple test for the API

## Types of changes


jchen-eb:
Nice addition, LGTM!

Issue #342, opened 2017-08-29
Don't erase custom properties from Bar Chart data set


piotrsucharski:
## Expected Behavior
Custom properties in dataset for barcharts ARE NOT ERASED and AVAILABLE on 'customMouseOver' as a callback argument.
## Current Behavior
Custom properties in dataset for barcharts ARE ERASED and NOT AVAILABLE on 'customMouseOver' as a callback argument.

## Possible Solution
Just handle the dataset the same way as you do for Donut chart without erasing additional properties from the dataset with cleanData method.

Use reduce instead of map to format the data you need in the way you like, but don't erase the custom properties.

```
function cleanData(data) {
            let cleanData = data.reduce((acc, d) => {
                // Skip data without quantity
                if (!d[quantityLabel]) {
                    return acc;
                }

                d.quantity = +d[quantityLabel];
                d.name = String(d[nameLabel]);
                d.percentage = d[percentageLabel] || null;
                acc.push(d);

                return acc;
            }, []);
            let totalQuantity = sumValues(cleanData);
            let dataWithPercentages = cleanData.map((d) => {
                d.percentage = String(d.percentage || calculatePercent(d[quantityLabel], totalQuantity, percentageFormat));

                return d;
            });

            return dataWithPercentages;
        }
```

## Steps to Reproduce

1. Build a Bar Chart with a dataset containing some custom properties eg.

```
var dataSet = [{
  value: 23,
  name: 'some label name',
  myCustomProp: 'my custom value'
 }, 
 ...
];

```
2. Add 'customMouseOver' event listener on Bar chart instance: 

```
this.barInstance.on('customMouseOver', (data) => {
      tooltip.show();
});
```

3. Check with debugging tools what data argument contains. It'll be: 
```
[{
 value: 23,
 name: 'some label name'
}]

```
4. myCustomProp was erased by cleanData method from bar.js: 

```
function cleanData(originalData) {
            let data = originalData.map((d) => ({
                    value: +d[valueLabel],
                    name: String(d[nameLabel])
                }));
            let dataZeroed = data.map((d) => ({
                    value: 0,
                    name: String(d[nameLabel])
                }));

            return {data, dataZeroed}
}

```
## Context
I need to add custom properties to dataset to access them on mouseover, but because of _cleanData method in bar.js all custom properties I've added are deleted. 

I'm performing the same operation on donut chart and there I'm completely free to use whatever properties I like as far as they fulfill the donut chart dataset requirements.

## Your Environment
[...]

PR #388, opened 2017-10-13
Updating cleanData functions all over the library

Golodhros:
## Description

We need to update the cleaning functions of the library so that they don't change the data entries and just cast the right type on the used values

## Motivation and Context

Some users showed their concerns at the data being cleaned when using events to recover it.
Related to https://github.com/eventbrite/britecharts/issues/342

## How Has This Been Tested?

Ran tests

## Types of changes


jchen-eb:
nice, LGTM

Issue #145, opened 2017-02-18
Add option to adjust the padding on the Bar Chart


Golodhros:
The padding between bars in the bar chart should be a design decision by our users. Let's make it configurable.

AlbaHoo:
+1

dalerasrorov-eb:
https://github.com/eventbrite/britecharts/pull/398

PR #396, opened 2017-10-16
Adds an option to adjust padding in barcharts

DalerAsrorov:
This change gives user an option to adjust the padding

## Description
Instead of having a fixed constant padding (0.1), I added a new function variable called `padding` that accepts a number and is used in `buildScales()` when adjusting the position of different bars in the chart. The default value remains 0.1 (hence it is passed in the module `padding = 0.1`). 

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/145

## How Has This Been Tested?
I added a new test in `test/specs/bar.spec.js` that checks whether the chart allows to set and get the padding. Initially the value is 0.1 (default value), so I pass it a different value (0.5) and see if the change was made accordingly based on expected and actual values given from `padding(number)` method.

screen shot 2017-10-16 at 1 27 11 am

## Types of changes

## Checklist:


docs/bar.js.html
```diff
@@ -270,6 +270,7 @@ <h1 class="page-title">Source: bar.js</h1>
             xTicks = 5,
             percentageAxisToMaxRatio = 1,
             enablePercentageLabels = false,
+            padding = 0.1,
```
miglesiasEB: Let's rename this to something a bit more specific, as it could be confused with a 'general' chart padding.

Maybe `betweenBarsPadding` or `barPadding`?

docs/bar.js.html
```diff
@@ -917,6 +918,22 @@ <h1 class="page-title">Source: bar.js</h1>
         };
 
         /**
+         * Gets or Sets the padding of the chart
+         * Default is 0.1
+         * @param  {Number} _x Padding value to get/set
+         * @return { padding | module} Current padding or Chart module to chain calls
```
miglesiasEB: This is wrong in a lot of our comments, it should be:
   @return { Number | module} Current....

Sorry for the confusion, we are fixing those as we see them!

Golodhros:
Oh, it seems you need to update master, rebase your branch and push again!

src/charts/bar.js
```diff
@@ -729,6 +730,22 @@ define(function(require) {
         };
 
         /**
+         * Gets or Sets the padding of the chart
+         * Default is 0.1
+         * @param  {Number | module} _x Padding value to get/set
+         * @return { padding | module} Current padding or Chart module to chain calls
```
Golodhros: @return {Number | module}

Issue #399, opened 2017-10-17
Tooltip render error on IE11


Golodhros:
## Expected Behavior

The tooltip should render properly

## Current Behavior

List of topics show up compressed.
![screen shot 2017-10-17 at 2 01 39 pm](https://user-images.githubusercontent.com/190833/31689529-880170c0-b344-11e7-82d1-6cb04a3bba77.png)

## Steps to Reproduce (for bugs)

1. Load the line chart demo in IE11
2. Check the tooltip in several points (sometimes it renders properly)

PR #401, opened 2017-10-18
Fix IE11 tooltip bug

Golodhros:
## Description

Fixes (more or less) the bug in the tooltip for IE11

## Motivation and Context

Fixing https://github.com/eventbrite/britecharts/issues/399

## How Has This Been Tested?

Tested in IE11

## Types of changes


jchen-eb:
LGTM

docs/tooltip.js.html
```diff
@@ -640,7 +657,7 @@ <h1 class="page-title">Source: tooltip.js</h1>
                     return -1;
                 });
 
-            let otherIndex = topics.map(({name}) => name).indexOf('Other');
+            let otherIndex = topics.map(({ name }) => name).indexOf('Other');
```
entobot: This line, and the lines following it, are unreachable; they'll never execute. This might be a bug. If it's not, I recommend you remove these lines to ensure code cleanliness.

Issue #303, opened 2017-07-10
Is it possible to choose the items order in Donut chart?


tonydtran:
Hi everyone,

I would like to show several items in the donut chart in a specific order.

Example:
`const data = [
  {
    quantity: 20,
    name: 'critical',
    id: 1
  },
  {
    quantity: 30,
    name: 'normal',
    id: 2
  },
  {
    quantity: 50,
    name: 'good',
    id: 3
  }
]` 

In this case, the "good" slice will be the first displayed in the donut in the clockwise rotation, then the "normal" slice and finally the "critical" slice, because briteCharts generates from the biggest slice to the smallest, right?

Is it possible to define my own order, like using the ids ? If it's the case, I didn't find how to do that in the documentation.

Sorry for my poor English and thank you for your help !

Golodhros:
Hi @tonydtran, thanks for the issue!

So you are right, by default we are ordering the slices by quantity. The function that we use is called 'sortComparator'. In order to make this configurable, we could create an accessor for this variable, so it would allow to change that function and change the ordering algorithm.

You could do it yourself and create a PR or if you can wait we could tackle this at some point in the near future.

tonydtran:
Actually I'm not skilled enough to do that so I can wait for an update. Maybe in few months if you didn't change that and I got enough skills, I would do that haha !

Thank you for your answer @Golodhros !

nobitagit:
Wouldn't it be easier to just show the data in the order that the chart receives it? This would have many advantages:

- the user would have total control on the data order, any sorting would be possible
- simpler API for the end user and simpler docs
- simpler maintenance of britecharts for contributors
- lighter footprint in terms of size (this pr would be about removing code, not adding)
- remove all the sorting logic from the chart(s)

Also, i think this would follow much of the concepts that we all use anyway when building stateless components in React, Vue and the likes where the goal is components that simply display data rather than changing it.

Golodhros:
Hi @nobitagit, 

Good call! Yeah, it makes sense from the point of view of the code. 

I will check with our designer @evb-sun, as at this point I am not sure about the context of the decision of ordering by slice size. I guess it will be OK.

Thanks!

Golodhros:
So, I checked with @evb-sun  and we think we will prefer to keep the default order as it is and add the option to change it.

The resultant API would be something like:
   donut.forceOrderingFunction(fn);

where the fn would be something like:
    (a, b) => b.quantity + a.quantity

or similar.

[...]

PR #409, opened 2017-10-23
Added ability to order items in donut chart

dalerasrorov-eb:
Add ability to customize the order of items in Donut chart. 

## Description
This PR adds the user the ability to add a custom function to manipulate the order of items displayed in donut chart by providing the function `orderingFunction(fn)`. This function is chain-able and can be invoked at different times after the donut chart setup is done. 

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/303

## How Has This Been Tested?
New tests were added to `donut.spec.js` to specifically test whether the order of the item was changed after the ordering function is invoked. There are two tests in total: 1) checks whether the default sorting function is invoked by testing the item index. 2) checks whether the order of items has changed accordingly (descending order) by checking index of the same node.  

You can see below that the order of items has reversed: 
screen shot 2017-10-20 at 9 11 56 pm
screen shot 2017-10-20 at 9 11 38 pm

## Types of changes

## Checklist:


docs/donut.js.html
```diff
@@ -290,15 +290,18 @@ <h1 class="page-title">Source: donut.js</h1>
...
             // utils
             storeAngle = function(d) {
                 this._current = d;
             },
             reduceOuterRadius = d => {
                 d.outerRadius = externalRadius - radiusHoverOffset;
             },
-            sortComparator = (a, b) => b.quantity - a.quantity,
+
+            // Default sortComparator: (a, b) => b.quantity - a.quantity,
```
Golodhros: Nit: you could get rid of this comment as it doesn't add much

docs/donut.js.html
```diff
@@ -717,6 +720,21 @@ <h1 class="page-title">Source: donut.js</h1>
         };
 
         /**
+         * Changes the order of items given custom function
+         * @param  {Function | Module} _x   A custom function that sets logic for ordering
```
Golodhros: The input will always be a Function if there is any

docs/donut.js.html
```diff
@@ -717,6 +720,21 @@ <h1 class="page-title">Source: donut.js</h1>
...
+         * @return { (void | Module) }      Void function with no return
```
Golodhros: The return I think will be Fucn | module

Issue #400, opened 2017-10-18
Drawing Axis Labels


Zangdarr:
For the purpose of a better chart reading, I would like to suggest the possibility to show labels for both the X and Y axes on the line chart (for my case, but can be extended to any type of chart).

## Expected Behavior
I thought about something like :
```
lineChart
    .setAxisLabelX("Production date")
    .setAxisLabelY("Product amount")
    .showAxisLabelX(true)
    .showAxisLabelY(true);
```
[https://imgur.com/a/VGOya](url)

## Current Behavior
Currently, I didn't find any way within the API to make it happen.

## Possible Solution
In britecharts/charts/lines.js
1. Add 4 variables : 
    - string axisLabelX
    - string axisLabelY
    - boolean showAxisLabelX
    - boolean showAxisLabelY

2. Add 4 functions : 
    - setAxisLabelX(string)
    - setAxisLabelY(string)
    - showAxisLabelX(boolean)
    - showAxisLabelY(boolean)

3. Update the function drawAxis() to draw the axes label if the corresponding flags are green.

Golodhros:
Hi @Zangdarr , thanks for the issue!

We have not had this necessity but I guess it will make sense for some of our users.

The API you propose is fine, although I would prefer to not use the booleans, as the presence of the label would be enough to determine if we should show it or not.

PR #410, opened 2017-10-25
Feature : Adding axis labels to line chart

Zangdarr:
## Description

I made it so users that create a line chart can specify labels for both axes X and Y and has it draw them on the chart. My changes need an intervention because I couldn't have the margins of the chart successfully increased to fit the two labels. Even though I had access to the margins const in the main function line(), updating it didn't achieve any purpose. The exports(_selection) function in line.js doesn't seem to consider the margins defined in the function line().

## Motivation and Context

Link to Issue #400

## How Has This Been Tested?
I followed the suggestion in the contributing guide of the project.
I updated the line.spec.js file to add four test to my features
- two accessors test, one for each label
- one to test the absence of the labels on the chart if the labels aren't specified (default behavior)
- one to test the presence of the labels on the chart if they are specified

I ran  : `Yarn run test` -> all tests are green

To illustrate the problem with the margins
Margins problem

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -139,6 +139,8 @@ define(function(require){
                 right: 0
             },
             monthAxisPadding = 28,
+            axisLabelPaddingX = 28,
+            axisLabelPaddingY = 28,
             tickPadding = 5,
             highlightCircleSize = 12,
             highlightCircleStroke = 2,
@@ -178,6 +180,8 @@ define(function(require){
             valueLabel = 'value',
             topicLabel = 'topic',
             topicNameLabel = 'topicName',
+            axisLabelX = null,
+            axisLabelY = null,
 
             yTicks = 5,
 
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -120,8 +120,8 @@ define(function(require){
         let margin = {
                 top: 60,
                 right: 30,
-                bottom: 40,
-                left: 70
+                bottom: 55, //40 -> 55 , modification of this margin to fit the axisLabelX but not working during the exports(_selections)
+                left: 85, //70 -> 85 , modification of this margin to fit the axisLabelY but not working during the exports(_selections)
             },
             width = 960,
             height = 500,
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -120,8 +120,8 @@ define(function(require){
         let margin = {
                 top: 60,
                 right: 30,
-                bottom: 55, //40 -> 55 , modification of this margin to fit the axisLabelX but not working during the exports(_selections)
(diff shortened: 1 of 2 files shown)


src/charts/line.js
```diff
@@ -139,6 +139,8 @@ define(function(require){
                 right: 0
             },
             monthAxisPadding = 28,
+            axisLabelPaddingX = 28,
```
Golodhros: This should probably be `xAxisLabelPadding`

src/charts/line.js
```diff
@@ -178,6 +180,8 @@ define(function(require){
             valueLabel = 'value',
             topicLabel = 'topic',
             topicNameLabel = 'topicName',
+            axisLabelX = null,
```
Golodhros: Let's do xAxisLabel and yAxisLabel instead

test/specs/line.spec.js
```diff
@@ -587,6 +587,90 @@ define([
...
+                    f.clearCache();
+                });
+
+                describe('on render', () => {
+
+                    it('should not render any axisLabel', () => {
+                        expect(containerFixture.selectAll('.axisLabel')["_groups"][0].length).toBe(0);
+                    });
+
+                    it('should provide axisLabelX getter and setter', () => {
```
Golodhros: I guess this tests are not dependent on the labels being set or not, so we usually place them with the rest of API tests

src/charts/line.js
```diff
@@ -524,12 +528,28 @@ define(function(require){
...
                 .transition()
                 .ease(ease)
                 .attr('transform', `translate(${-xAxisPadding.left}, 0)`)
                 .call(yAxis)
                 .call(adjustYTickLabels);
+
+            if (axisLabelY) {
+                svg.select('.chart-group')
+                    .append("text")
+                    .attr("class", "axisLabel")
```
Golodhros: I will name the class .y-axis-label

src/charts/line.js
```diff
@@ -524,12 +528,28 @@ define(function(require){
                     .call(xMonthAxis);
             }
 
+            if (axisLabelX) {
+                svg.select('.x-axis-group')
+                    .append("text")
+                    .attr("class", "axisLabel")
```
Golodhros: I would class it with '.x-axis-label'

Zangdarr:
The suggested changes have been implemented. Thank you for the return

Golodhros:
Awesome! Thanks a lot!

I will try to find some time this weekend and fix the issue, so that we can merge it.

Golodhros:
Merged in #419

PR #413, opened 2017-10-26
Base tick number calculation on difference between minDate and maxDate instead of array length

jchen-eb:
## Description
The tick calculation doesn't work well when you have data that skips days, since it is based on array length rather than the difference in days between the minDate and maxDate.

For example, I have a data set that spans from Jan 5 to Feb 6, but there are lots of "holes" in the data so we only have 12 "days" within that range that have data attached.

However it's actually 32 days, but the tick count function believes it's 12 days, causing way too many ticks to be rendered.

## How Has This Been Tested?

Change unit tests

**Before fix:**
screen shot 2017-10-26 at 12 04 18 pm
small screen:
screen shot 2017-10-26 at 12 35 05 pm

**After fix:**
screen shot 2017-10-26 at 12 02 32 pm
small screen:
screen shot 2017-10-26 at 12 03 06 pm

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/helpers/timeAxis.js
+++ b/src/charts/helpers/timeAxis.js
@@ -86,6 +86,13 @@ define(function(require) {
     }
 
     /**
+     * Takes a number representing milliseconds and convert to days
+     * @param  {Number} milliseconds    Any number
+     * @return {Number}                 Number of days that the input represents
+     */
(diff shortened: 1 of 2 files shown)


miglesiasEB:
LGTM

PR #416, opened 2017-11-01
Fixed the wrong comment of the method description

dalerasrorov-eb:
Fixing the error I made in last commit in docs. Wrong description was added, so fixed it... My bad!


Golodhros:
LGTM

Hahaha, I loved your last commit message.
Thanks a lot!

Issue #412, opened 2017-10-26
Sparkline failing to style gradients when multiple instances


Golodhros:
## Expected Behavior

We should be able to render multiple sparklines in one page

## Current Behavior

The styles are not being applied, but the paths are there

## Possible Solution

Remove the ids like #sparkline-line-gradient

## Steps to Reproduce (for bugs)

1. Check https://codepen.io/Britecharts/pen/VModxX?editors=1010#0

PR #421, opened 2017-11-06
Fix sparkline multiple instances (#412)

amber-eb:
## Description
Fix a bug where the same gradient / animation settings applied to multiple instances of the Sparkline chart. Also fixes a similar bug found in the Brush chart.

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/412

## How Has This Been Tested?
Ran this code in the "demo" section of my local britecharts docs site.
https://codepen.io/Britecharts/pen/VModxX?editors=1010#0

## Types of changes

## Checklist:

- [x] I have updated existing tests to cover my changes


src/charts/helpers/common.js
```diff
@@ -2,6 +2,7 @@ define(function(require) {
     'use strict';
 
     const d3Format = require('d3-format');
+    let idCounter = 0;
```
jchen-eb: It would be better if we could generate `uniqueIds` without any side effects like mutating a counter stored in this scope. Maybe we could use this module? https://www.npmjs.com/package/uuid

I think we do use this approach in some places so it's not that big of a deal but I think it's better in general to avoid this if you can.

amber-eb: Would you mind clarifying why you recommend against this approach? I understand that we should avoid side effects that mutate data in unexpected ways, but `idCounter` exists for exactly this purpose. This implementation was more or less grabbed from the [`_.uniqueId`](https://github.com/lodash/lodash/blob/4.0.1-npm-packages/lodash.uniqueid/index.js) source code.

I do see some potential issues with this approach: for example, this way of incrementing IDs might give the false impression that IDs are being applied in a deterministic manner (since the ID counter increments each time the function is called instead of being apparently random). That said, using a long UUID for an SVG element ID would make the resulting markup much less readable IMO.

miglesiasEB: IMHO this will be fine for now.
The uuid package is definitely more robust and professional, but I think we could save some KB by not using it.

src/styles/charts/brush.scss
```diff
@@ -12,7 +12,6 @@
         // stroke-width: 1;
         fill-opacity: 0.08;
         // fill-opacity: 1;
```
miglesiasEB: Let's remove all this commented styles too, thanks!

src/charts/helpers/common.js
jchen-eb: As a general rule I think modules should be as pure and stateless as possible. This is to avoid any unpredictability with the usage of the module, for example if you are importing this module in N number of places, it can be difficult to tell which importer will get "1", who will get "2" or "3", etc. And the answer will change if you have multiple instances of the module in your environment for some reason (you shouldn't, but IME dependency systems can sometimes do things you don't expect). The first time you import from instance 1 you'll get "1", the first time you import from instance 2 you''ll get "1". Does it matter? Probably not, but at the same time I think storing state in a module should be the exception not the rule, and if you don't *need* to do it you should avoid it.

The `uuid` module is tiny, and even if we had a problem with its size you can implement a uuid-generating function in one short line of code: https://gist.github.com/jed/982883

It's definitely not a big deal, and I have a hard time coming up with scenarios where this would ever be an issue. You're right about the shorter SVG ID resulting in nicer markup, and maybe that is enough of a benefit that we don't have to force ourselves to use pure and stateless modules for the sake of having them be pure and stateless. I'm fine either way :)

jchen85: Although if lodash is using it then it's probably not an issue at all :)

test/specs/sparkline.spec.js
```diff
@@ -24,6 +24,10 @@ define([
         return _.contains(element[0][0].classList, className);
     };
 
+    const hasIdWithPrefix = (element, prefix) => {
+        return element['id'].includes(prefix);
```
prodobot: You do not need to use a member expression `element['id']` to access this property - you can instead use dot notation as follows:
```javascript
  return element.id.includes(prefix);
```

src/charts/helpers/common.js
amber-eb: @jchen-eb thanks for the explanation! Let's stick with this for now but keep `uuid` on our radar in case some issue does arise with this implementation.

jchen-eb: Sounds good! 👍

Issue #424, opened 2017-11-09
Line chart y-axis labels don't render when chart loads in a hidden document / background tab


CoryDuncan:
The problem appears to be the `transition()` call in https://github.com/eventbrite/britecharts/blob/master/src/charts/line.js#L552-L553. D3 runs `transition()` in a `requestAnimationFrame()` and in most browsers, calls to `requestAnimationFrame()` are paused for hidden documents. In this case, it seems to prevent the line chart y-axis labels from rendering. Tested using Chrome 61.

## Expected Behavior
Line chart y-axis labels are rendered when chart loads regardless of document visibility.

## Current Behavior
Line chart y-axis labels are not rendered when chart loads while the document is hidden.

## Steps to Reproduce (for bugs)
With a line chart containing y-axis labels, load the chart while the document is hidden.

Golodhros:
As mentioned in the PR, let's get rid of that transition.

PR #425, opened 2017-11-09
Fix disappearing line chart y-axis labels when chart is loaded in hidden document

CoryDuncan:
Fixes https://github.com/eventbrite/britecharts/issues/424

## Description
`transition().ease()` is now only called on the line chart y-axis when the document is not hidden.


Golodhros:
Hi @CoryDuncan , thanks for the Issue report and the fix!

We have talked about this, and we actually think that the best option is to actually get rid of that transition. This was something that we don't even want from a design point of view, so simply by removing:

  .transition()
  .ease(ease)

It would be enough and better for the future.

Thanks a lot!

CoryDuncan:
@Golodhros I agree, from a design perspective, it feels better without the transition. I'll close this and open a separate PR to remove it.

PR #426, opened 2017-11-10
Remove transition from line chart y-axis

CoryDuncan:
Fixes https://github.com/eventbrite/britecharts/issues/424
Derived from https://github.com/eventbrite/britecharts/pull/425

## Description
This removes `transition().ease()` from the line chart y-axis.

### DIFF
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -549,8 +549,6 @@ define(function(require){
             }
 
             svg.select('.y-axis-group .axis.y')
-                .transition()
-                .ease(ease)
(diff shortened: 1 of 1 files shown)


Golodhros:
LGTM

Thanks a lot for your contribution!

BTW, it is not necessary to run the build, we will do it before releasing.

CoryDuncan:
> BTW, it is not necessary to run the build, we will do it before releasing.

Thanks, I was unsure about that. I've updated the PR accordingly.

miglesiasEB:
Thanks a lot!

Issue #408, opened 2017-10-20
Add Active effect on highlight points


miglesiasEB:
## Expected Behavior

We should show an 'active' state when somebody clicks in one of the highlight points on the line and stacked bar charts

## Current Behavior

Nothing happens

![fullscreen_10_20_17__3_13_pm](https://user-images.githubusercontent.com/9063187/31843757-9039c4c4-b5a9-11e7-8c58-d1b6229f918e.png)

## Possible Solution

Add an svg filter with a blur of the proper color when we click on them

miglesiasEB:
Check some ideas in  http://4waisenkinder.de/blog/2013/09/28/using-gradient-and-shadows-with-d3-dot-js/

PR #430, opened 2017-11-19
Adds glow on 'active' state to the highlight points in Line and Area charts

Golodhros:
## Description

Adding a svg filter with glow to the highlight points on the line and area charts

## Motivation and Context

Fixes https://github.com/eventbrite/britecharts/issues/408

## How Has This Been Tested?

Ran tests

up3tspswyc

## Types of changes

### DIFF
--- a/src/charts/stacked-area.js
+++ b/src/charts/stacked-area.js
@@ -26,7 +26,7 @@ define(function(require){
     } = require('./helpers/formatHelpers');
     const {
         createFilterContainer,
-        createGlow,
+        createGlowWithMatrix,
     } = require('./helpers/filters');
     const {
         isInteger,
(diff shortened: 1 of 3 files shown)


src/charts/stacked-area.js
```diff
@@ -941,18 +962,22 @@ define(function(require){
...
-                              .append('g')
-                                .classed('circle-container', true)
-                                  .append('circle')
-                                    .classed('data-point-highlighter', true)
-                                    .attr('cx', highlightCircleSize)
-                                    .attr('cy', 0)
-                                    .attr('r', highlightCircleRadius)
-                                    .style('stroke-width', highlightCircleStroke)
-                                    .style('stroke', categoryColorMap[d.name])
-                                    .on('touchstart click', function() {
```
jchen-eb: will removing the `touchstart` event affect functionality for mobile users?

miglesiasEB: I don't think so, it was a security thing but I don't think is necessary anymore. 
Will check though

Issue #350, opened 2017-09-11
Enable number formatting on all chart types.


piotrsucharski:
I'd propose to add numberFormat() method on all chart types not only on Legend. As I see you are using the numberFormat variable on all chart/tooltip types, but it's exposed only on Legend type. I don't see any reason for not exposing it for other types.

## Expected Behavior
Formatting a number is possible for all chart/tooltip types where numberFormat is used.

## Current Behavior
Formatting a number is possible only for Legend type.

## Possible Solution
It should be as simple as adding numberFromat method to exports as it is on Legend.

```
 exports.numberFormat = function(_x) {
            if (!arguments.length) {
                return numberFormat;
            }
            numberFormat = _x;

            return this;
        };
```

## Context
I use charts only for showing full numbers without any 0.5, 0.23 etc. I'd like to present number 4 to the user as just 4 not 4.00.

## Your Environment

* Version used: 2.0.0
* Browser Name and version: Chrome 60
* Operating System and version (desktop or mobile): Win 7

Golodhros:
Hi @piotrsucharski , thanks for the issue!

That seems doable. There is a bunch of parameters that we haven't make open because we didn't know if they would be, so it makes sense for us to make them available as users like you need them.

Out of curiosity, in which specific charts would you need this?

Thanks!

piotrsucharski:
I am using the following components:
- Bar
- Donut
- Legend
- Mini Tooltip
- Tooltip
- Brush
- Line

Golodhros:
Thanks!

PR #433, opened 2017-11-26
Feature - donut number formatting

dalerasrorov-eb:
## Description
Added a new property called `numberFormat`. This function will reformat the value of the data item in donut chart depending on the string format given to the `numberFormat(format)` function.

## Motivation and Context
Another addition to https://github.com/eventbrite/britecharts/issues/350

## How Has This Been Tested?
Added two new tests two `donut.spec.js`. If `numberFormat` is not given, then it's not going to format the value. This was done to not fail the previous tests. The `numberFormat` is changed only this function is applied.

screen shot 2017-11-26 at 12 07 57 am

## Types of changes

## Checklist:


src/charts/donut.js
```diff
@@ -651,6 +659,21 @@ define(function(require) {
             return value === dispatcher ? exports : value;
         };
 
+        /**
+         * Changes the order of items given custom function
+         * @param  {Function} _x              A custom function that sets logic for ordering
+         * @return { (Function | Module) }    Void function with no return
+         * @public
+         */
+        exports.orderingFunction = function(_x) {
```
Golodhros: Thanks for the re-ordering!!

PR #441, opened 2017-12-08
Feat donut empty

amber-eb:
Adds an empty state for the donut chart. For backwards compatibility, it defaults to false.

## Description
Renders an empty slice with a piece of spoofed data. Keeps the other slices around so that they can be shown as highlighted slices in the tooltip.

## Motivation and Context
We need an empty donut chart for a spec.

## How Has This Been Tested?
I added tests to ensure that in both cases where data is an empty array and when data is an array of objects each with quantity zero, if the empty state is configured on, an empty slice will be rendered. The tests also check the the custom empty state color applies in each of these cases.

## Types of changes

## Checklist:

### DIFF
--- a/demos/donut.html
+++ b/demos/donut.html
@@ -87,11 +87,10 @@ donutContainer.datum(dataset).call(donutChart);
 
 <div class="container">
     <div class="row">
-        <div class="
-col-md-6">
+        <div class="col-md-6">
             <section>
                 <article>
-                    <h2>Donut Chart with fixed higlighted slice</h2>
+                    <h2>Donut Chart with fixed highlighted slice</h2>
                     <div class="js-donut-highlight-slice-chart-container card--chart"></div>
                 </article>
             </section>
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -14,6 +14,7 @@ define(function(require) {
     const textHelper = require('./helpers/text');
     const colorHelper = require('./helpers/colors');
     const {calculatePercent} = require('./helpers/common');
+    const {emptyDonutData} =require('./helpers/constants');
 
 
     /**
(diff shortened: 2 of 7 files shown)


demos/donut.html
```diff
@@ -87,11 +87,10 @@ <h4>Demo Code</h4>
 
 <div class="container">
     <div class="row">
-        <div class="
-col-md-6">
+        <div class="col-md-6">
             <section>
                 <article>
-                    <h2>Donut Chart with fixed higlighted slice</h2>
+                    <h2>Donut Chart with fixed highlighted slice</h2>
```
Golodhros: Thanks!

src/charts/donut.js
```diff
@@ -511,6 +550,21 @@ define(function(require) {
             return this;
         };
 
+        /**
+         * Gets or Sets the emptyDataConfig of the chart
+         * @param  {Object} _x emptyDataConfig object to get/set
+         * @return { Object | module} Current config for when chart data is an empty array
+         * @public
```
Golodhros: Let's add an example here

test/specs/donut.spec.js
```diff
@@ -111,6 +111,125 @@ define(['d3', 'donut', 'donutChartDataBuilder'], function(d3, chart, dataBuilder
...
+                    let expected = '#000000';
+
+                    let emptyDataConfig = {
+                        emptySliceColor: expected,
+                        showEmptySlice: true,
+                    }
+
+                    donutChart.emptyDataConfig(emptyDataConfig);
+
+                    let dataset = [];
```
Golodhros: I would just inline this

Golodhros:
LGTM

PR #443, opened 2017-12-11
Expose additional donut configs

amber-eb:
Add functions to donut to allow client to set `radiusHoverOffset` and `percentageFormat`.

## Motivation and Context
We would like to be able to customize these properties in the invites module.

## How Has This Been Tested?
Manual testing + running existing tests.

## Types of changes

## Checklist:


src/charts/donut.js
```diff
@@ -301,6 +301,10 @@ define(function(require) {
             newSlices.merge(slices)
                 .attr('fill', emptyDataConfig.emptySliceColor)
                 .attr('d', shape)
+                .transition()
+                .ease(ease)
+                .duration(pieDrawingTransitionDuration)
+                .attrTween('d', tweenLoading);
```
amber-eb: I also snuck in this change to animate the empty slice state.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "britecharts",
-  "version": "2.4.7",
+  "version": "2.4.8",
```
jchen-eb: You don't need to update the version here (or in bower.json) since our release script will do it automatically when it's run.

amber-eb: Got it!

PR #479, opened 2018-01-09
Updating changelog

Golodhros:
Updating Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,47 @@
+
 2.4.11
-2.4.10
+--------------
+Features:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,18 +1,47 @@
...
+--------------
+Bug fixes:
+* Defensive check on text helper
+
 2.4.2
-2.4.1
-2.4.0
+--------------
+Features:
+* Docs build on release
```
jchen-eb: This one is repeated twice (also in 2.4.7)

We could probably remove it since it's not relevant to users.

Issue #481, opened 2018-01-11
Add .legendTextFunction to Donut Chart


mrbongiolo:
## Expected Behavior
Add a new `legendTextFunction` method on the API.

## Current Behavior
Right now `drawLegend` calls `text( () => '${obj.data.percentage}% ${obj.data.name}' )`.

## Possible Solution
The new method `legendTextFunction` should be used on `drawLegend`, it will be called with `obj.data`, so that the custom function can have access to the data.
By default it should display the same text that we have today.

## Context
Allowing the usage of a custom function will help in cases when someone doesn't want to show the `name` or might want to use `quantity` instead of the `percentage`, alongside other options.
By following the concepts of "convention over configuration" we should always provide a good default behavior, in this case the same "45% Something" that we have today, so that user can start using the lib easily, but allowing for different and complex scenarios by providing a good API to hook in custom methods.

Golodhros:
This looks great @mrbongiolo, thanks for creating the issue!

I have only one comment. We called that text legend, but now in hindsight I think is a poor naming, especially as we have a Legend component. Basically, I would love to move out from that name, and at the same time I don't know how we should call it.

Maybe 'centerText'? Or 'centerLabel'? Any suggestion will be appreciated.

mrbongiolo:
@Golodhros Yeah, 'text legend' might be confusing.

In this case I prefer 'centerText' over 'centerLabel', since 'label' is used on the data as well and it could be confusing.

TBH, since I started using the lib the `nameLabel` and `valueLabel` methods always seemed a bit odd, they aren't quite right in point on what they do, without the docs at first I thought they were some kind of label element.

Wouldn't it be better to be "stricter" with the data schema? A BarChart should always have at least the `name` and `value` fields (any other fields should use other names and would be accessible in some functions, etc) , it's better to transform the data before sending it to the chart, than naming the `name` field as `something` and then ask the chart to use `something`.
This would allow the usage of the term `label` like an element. What do you think? I know this would require some changes and a new version that isn't backwards compatible, but it would make things clearer.

Golodhros:
@mrbongiolo that's a good point about the 'nameLabel' and 'valueLabel'. The presence of those options in the API is something that since the beginning I have doubted.

In my mind, from the point of view of engineering and API correctness, they are almost an aberration. The data schema should be fixed, and our users would need to do the data transformations they need to follow that schema.

[...]

PR #498, opened 2018-01-31
Feature - added text formatter function to Donut's API

dalerasrorov-eb:
## Description
This feature allows user to customize the format of the text shown in the middle of the chart on hover. If the user wants to show quantity instead of percentage, this function can be used to do that.

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/481

## How Has This Been Tested?
Added new tests to `donut.spec.js` that checks:
* whether the default function is properly applied and correct value is returned
* if the function is applied, validate that the new format values are matching the expected 

Screenshot of the format that takes in `quantity` as a value and `name` as label.
centeredtextformat

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -558,6 +558,13 @@ define(function(require) {
 
         // API
 
+        /**
+         * Gets or Sets the centeredTextFunction of the chart. If function is provided
+         * the format will follow the custom function's value format.
+         * @param  {Function} _x        Custom function that returns a formatted string
+         * @return {Function | module}  Current centeredTextFunction or Chart module to chain calls
+         * @public
+         */
         exports.centeredTextFunction = function(_x) {
             if (!arguments.length) {
                 return centeredTextFunction;
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -561,6 +561,8 @@ define(function(require) {
         /**
          * Gets or Sets the centeredTextFunction of the chart. If function is provided
          * the format will be changed by the custom function's value format.
+         * The default format function value is "${d.percentage}% ${d.name}"
+         * The callback will provide the data object with id, name, percentage, and quantity
          * @param  {Function} _x        Custom function that returns a formatted string
          * @return {Function | module}  Current centeredTextFunction or Chart module to chain calls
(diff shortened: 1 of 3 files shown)


src/charts/donut.js
```diff
@@ -556,6 +558,22 @@ define(function(require) {
 
         // API
 
+        /**
+         * Gets or Sets the centeredTextFunction of the chart. If function is provided
+         * the format will be changed by the custom function's value format.
+         * @param  {Function} _x        Custom function that returns a formatted string
+         * @return {Function | module}  Current centeredTextFunction or Chart module to chain calls
+         * @public
```
Golodhros: I would add here some more info about how to use this, like the default format and the data that the user could expect to have within the 'd' object.

test/specs/donut.spec.js
```diff
@@ -290,7 +290,51 @@ define(['d3', 'donut', 'donutChartDataBuilder'], function(d3, chart, dataBuilder
...
+                    let valueNode = containerFixture.select('text.donut-text .value').nodes()[0];
+                    let labelNode = containerFixture.select('text.donut-text .label').nodes()[0];
+
+                    actualValue = d3.select(valueNode).text();
+                    actualLabel = d3.select(labelNode).text();
+
+                    expect(actualValue).toBe(expectedValue);
+                    expect(actualLabel).toBe(expectedLabel);
+                });
+            });
```
Golodhros: Awesome!

src/charts/donut.js
```diff
@@ -556,6 +558,24 @@ define(function(require) {
 
         // API
 
+        /**
+         * Gets or Sets the centeredTextFunction of the chart. If function is provided
+         * the format will be changed by the custom function's value format.
+         * The default format function value is "${d.percentage}% ${d.name}"
+         * The callback will provide the data object with id, name, percentage, and quantity
```
miglesiasEB: I would add that we also provide whatever components the user adds in each data entry.

I would also add an example, like:
```
@example
 donutChart
     .centeredTextFunction((d) => `${d.id} ${d.quantity}`);
```

PR #499, opened 2018-01-31
Feat stacks reverse

sound-matt:
Add an option to reverse stack order

## Motivation and Context

In vertical mode reverse stacks allow to have the tooltip legend in the same order as the stacks of the charts
In horizontal mode it's better to have stacks in default order.
I didn't want to change the default behavior & change stacks order for everyone, maybe someone code something to have a precise order for stacks

## How Has This Been Tested?

yarn test -> ok
Works with the demos

35743261-59006ada-083d-11e8-8c3f-bca568c2b62d

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/stacked-bar.js
+++ b/src/charts/stacked-bar.js
@@ -111,7 +111,7 @@ define(function(require){
             transformedData,
             stacks,
             layerElements,
-            reverseStacks = false,
+            hasReversedStacks = false,
 
             tooltipThreshold = 480,
 
@@ -694,7 +694,7 @@ define(function(require){
         function prepareData(data) {
             stacks = uniq(data.map(({stack}) => stack));
 
-            if (reverseStacks) {
+            if (hasReversedStacks) {
                 stacks = stacks.reverse();
             }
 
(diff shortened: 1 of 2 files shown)


dalerasrorov-eb:
Can you paste in a screenshot of demo?

miglesiasEB:
Maybe the reversed order should be the default?

in case we want to go with this option, we should prefix the option with 'is, has, can...' as it is a boolean.

dalerasrorov-eb:
Yes, coming from @Golodhros suggestion, `hasReversedStacks` would make sense.

src/charts/stacked-bar.js
```diff
@@ -867,6 +873,21 @@ define(function(require){
             return this;
         };
 
+        /**
+        * Gets or Sets the hasReversedStacks property of the chart, it reverse the stacks order.
```
DalerAsrorov: ```Gets or Sets the hasReversedStacks property of the chart, reversing the order of stacks.```

src/charts/stacked-bar.js
```diff
@@ -867,6 +873,21 @@ define(function(require){
...
+         * @param  {boolean} _x Desired horizontal direction for the graph
+         * @return { hasReversedStacks | module} If it is horizontal or Bar Chart module to chain calls
```
DalerAsrorov: Instead of 
```
@return { hasReversedStacks | module} If it is horizontal or Bar Chart module to chain calls
```
I would stick with the format used in the rest of the file which is:
```
@return { hasReversedStacks | module } Current hasReversedStacks or Chart module to chain calls
```
It is good to mention that this returns the value that tells whether the stacks are in horizontal above `@param`

sound-matt: Oops I missed that, I'll fix that

sound-matt:
Here is the screenshot of demo

britecharts hasReversedStacks

Issue #485, opened 2018-01-16
Disable bar chart darken on hover


chinaowl:
Question: is it possible to disable the bar chart darkening on hover? I actually want the opposite effect, where the bar that's being hovered over stays the same but the other bars get darker, which I can do with a custom mouseover event. But it seems like the darkening will always happen because it comes after (in the source code):

```
function handleMouseOver(e, d, chartWidth, chartHeight) {
    dispatcher.call('customMouseOver', e, d, d3Selection.mouse(e), [chartWidth, chartHeight]);
    d3Selection.select(e).attr('fill', ({name}) => d3Color.color(colorMap(name)).darker());
}
```

EDIT: I've come up with a workaround that _mostly_ works. On the first hover, all bars darken (haven't figured out why), but on subsequent hovers, it has the effect that I want. https://codepen.io/chinaowl/pen/jYejrd

Golodhros:
Hi @chinaowl, thanks for opening an issue!
I do agree with you, this shouldn't be a fixed behavior but an optional one.

I would suggest to basically remove that line that darkens the bar color and the one that restores it's color.

Thanks!

dalerasrorov-eb:
@chinaowl, I implemented something that might be handy https://github.com/eventbrite/britecharts/pull/500

chinaowl:
@dalerasrorov-eb awesome, thank you!

dalerasrorov-eb:
@chinaowl this weekend or next Monday!

dalerasrorov-eb:
Closing this @chinaowl, the functionality was released in `2.5.0` and the new functionality to apply a custom behavior to highlighted bars is ready https://github.com/eventbrite/britecharts/pull/503 and will be released later. Thanks for opening the issue!

PR #500, opened 2018-02-01
Feature - allow ability to switch Bar's hover behavior

dalerasrorov-eb:
Added a new feature that allows the user to be able to set the `hover` behavior over bar chart. 

## Description
Added a new function to `Bar`'s API called `hasSingleHover` which accepts a boolean value. 
* If `hasSingleHover` is not called (the default value is `true`), the behavior of the hover mode behaves as default: if one hovers over a bar, the hovered bar is darkened while the rest stay the same.
* If `hasSingleHover` is set to `false`, on hovering the bar all the bars but the hovered one will darken instead (the opposite).

## Motivation and Context
This feature will allow the user to choose between different bar hover behavior based on the one that makes sense. I saw this issue https://github.com/eventbrite/britecharts/issues/485 which gave me an idea of making it configurable.

## How Has This Been Tested?
Added 3 new tests to `bar.spec.js`:
1) One that checks setter and getter of `hasSingleHover`
2) Checks if the default behavior (color switch on hover) stays the same as before.
3) When `hasSingleHover(false)` check that the color of the hovered bar stays the same (since others are darkened).

* `hasSingleHover(false)` - only the hovered bar kept its color while others are darkened
togglehoverdark
*  `hasSingleHover()` - by default returns `true`, in this case the behavior is the same as before.
screen shot 2018-01-31 at 8 58 40 pm

## Types of changes

## Checklist:

**Update:** Renamed function to `hasSingleBarHighlight`

### DIFF
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -765,7 +765,7 @@ define(function(require) {
          * If the value is true (default), the only the hovered bar
          * will switch to darker color. If the value is false, only the hovered bar will stay
          * the same while the rest of the bars will switch to darker colors.
-         * @param  {boolean} _x        Should darken the hovered bar
+         * @param  {boolean} _x        Should highlight the hovered bar
          * @return { boolean | module} Is hasSingleBarHighlight used or Chart module to chain calls
          * @public
          */
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -98,6 +98,7 @@ define(function(require) {
             isHorizontal = false,
             svg,
 
+            hasSingleHover = true,
             isAnimated = false,
             ease = d3Ease.easeQuadInOut,
             animationDuration = 800,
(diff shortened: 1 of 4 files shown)


Golodhros:
The implementation is great, and it completely solves #485, but I am not sure if this is what we want for this chart.

Specifically, I am worried about restricting our options here. The current implementation is restrictive as it doesn't allow the user to choose what to do with the hovered bar, we just darken it and that's it. If the user wanted to lighten it, fade it or something else, he won't be able to do it.

Thinking out loud here, but if we just get rid of the darkening action, wouldn't the user be able to do whatever he wants just with d3 selections? Maybe is a bit complicated, but given that we provide the hovered bar in 'e', probably we can select all the bars, remove the one passed here and darken them, basically doing what you did in the lines 649-654.

We could argue that this is right, but too complex to do for our users, given that is such a usual thing to do.

Maybe another solution would be to keep what you are proposing in this PR, but creating a function that deals with the styling of the selected bar. By default it would be darkening it, but it could be something different. This solution won't break the current functionality.

Let me state this solution better, it would mean:
- Change hasSingleHover into 'hasSingleBarHighlight' that is true as default
- Create a 'highlightFunction' that, for a passed in bar, it applies a darkening by default. We would use this as well when 'hasSingleBarHighlight' is false

What do you think?

DalerAsrorov:
I agree that this function limits the user to only do the hover effect without passing the data point. I think that this function can be used for highlighting fast and I like the second solution that you proposed which is to change `hasSingleHover` to `hasSingleBarHighlight` that is set to `true` by default and create new function that will allow user to perform more complex operations on a bar within passed callback.

As I understand, `hasSingleBarHighlight` will be a function for quick solution and for those who might be less "tech-savy" while `highlightFunction` will give an option to customize it. That sounds good to me.

Golodhros:
Yeah, awesome!

Thanks a lot for your late push on Britecharts, you are on fire!! 🔥

src/charts/bar.js
```diff
@@ -761,19 +761,19 @@ define(function(require) {
...
 
         /**
-         * Gets or Sets the hasSingleHover status.
+         * Gets or Sets the hasSingleBarHighlight status.
          * If the value is true (default), the only the hovered bar
          * will switch to darker color. If the value is false, only the hovered bar will stay
          * the same while the rest of the bars will switch to darker colors.
-         * @param  {boolean} _x     Should darken the hovered bar
-         * @return { boolean | module} Is hasSingleHover used or Chart module to chain calls
+         * @param  {boolean} _x        Should darken the hovered bar
```
Golodhros: Nit: I would say 'highlight' instead of 'darken', in preparation for the next change

PR #503, opened 2018-02-04
Feature - new Bar chart API method to accept a custom function for highlighted bars

dalerasrorov-eb:
## Description
Added a function to `Bar` chart's API  to allow the user to pass a function as a callback to `highlightBarFunction` which which passes the highlighted bar through the callback for the user to manipulate it.. That way, developer can operate on a bar. For example, change the `fill` property using select or `d3Color`.

```
barChart.highlightBarFunction(barSelection => barSelection.attr('fill', 'blue'))
```

## Motivation and Context
There was initial discussion in https://github.com/eventbrite/britecharts/issues/485 and in the merged PR: https://github.com/eventbrite/britecharts/pull/500 on allowing the user to have more customization over `Bar` chart highlighting.

## How Has This Been Tested?
* Added 2 new tests to `bar.spec.js` that check whether the highlighting is applied with custom function correctly when `hasSingleBarHighlight` is either set to `true` or `false`.
* `yarn test`.

Applying `lightblue` color `barSelection.attr('fill',  'lightblue')`:
* `hasSingleBarHighlight` is `true` (Default)
screen shot 2018-02-03 at 11 33 56 pm

* `hasSingleBarHighlight` is `false`:
screen shot 2018-02-04 at 12 57 20 am

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -765,9 +765,10 @@ define(function(require) {
         /**
          * Gets or Sets the hasSingleBarHighlight status.
          * If the value is true (default), only the hovered bar is considered to
-         * be highlighted and will be switch to darker color. If the value is false,
+         * be highlighted and will be darkened by default. If the value is false,
          * all the bars but the hovered bar are considered to be highlighted
-         * and will switch to darker color.
+         * and will be darkened (by default). To customize the bar highlight or
+         * remove it completely, use highlightBarFunction instead.
          * @param  {boolean} _x        Should highlight the hovered bar
          * @return {boolean | module} Is hasSingleBarHighlight used or Chart module to chain calls
          * @public
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -763,9 +763,10 @@ define(function(require) {
 
         /**
          * Gets or Sets the hasSingleBarHighlight status.
-         * If the value is true (default), the only the hovered bar
-         * will switch to darker color. If the value is false, only the hovered bar will stay
-         * the same while the rest of the bars will switch to darker colors.
+         * If the value is true (default), only the hovered bar is considered to
+         * be highlighted and will be switch to darker color. If the value is false,
+         * all the bars but the hovered bar are considered to be highlighted
+         * and will switch to darker color.
          * @param  {boolean} _x        Should highlight the hovered bar
          * @return { boolean | module} Is hasSingleBarHighlight used or Chart module to chain calls
          * @public
@@ -806,6 +807,7 @@ define(function(require) {
          * @return {highlightBarFunction | module} Is highlightBarFunction used or Chart module to chain calls
          * @public
          * @example barChart.highlightBarFunction(e => d3.select(e).attr('fill', expectedHighlightColor))
+         * @example barChart.highlightBarFunction(e => null) // will disable the default highlight effect
          */
         exports.highlightBarFunction = function(_x) {
             if (!highlightBarFunction.length) {
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -105,7 +105,7 @@ define(function(require) {
             animationStepRatio = 70,
             interBarDelay = (d, i) => animationStepRatio * i,
 
-            highlightBarFunction = (bar) => d3Selection.select(bar).attr('fill', ({name}) => d3Color.color(colorMap(name)).darker()),
+            highlightBarFunction = (barSelection) => barSelection.attr('fill', ({name}) => d3Color.color(colorMap(name)).darker()),
             orderingFunction,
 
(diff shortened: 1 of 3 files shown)


src/charts/bar.js
```diff
@@ -793,12 +795,35 @@ define(function(require) {
...
+
+            return this;
+        }
+
         /**
          * Gets or Sets the isAnimated property of the chart, making it to animate when render.
          * By default this is 'false'
          *
          * @param  {Boolean} _x Desired animation flag
-         * @return { isAnimated | module} Current isAnimated flag or Chart module
```
DalerAsrorov: Thought it would be a good idea to remove those spaces, seen on several occasions

src/charts/bar.js
```diff
@@ -642,15 +643,15 @@ define(function(require) {
...
+                highlightBarFunction(e);
                 return;
             }
 
             barList.forEach(barRect => {
                 if (barRect === e) {
                     return;
                 }
-                d3Selection.select(barRect).attr('fill', ({name}) => d3Color.color(colorMap(name)).darker());
+                highlightBarFunction(barRect);
```
amber-eb: Nice! Added benefit of making this code more readable.

src/charts/bar.js
```diff
@@ -961,7 +986,7 @@ define(function(require) {
         /**
          * Gets or Sets whether the color list should be reversed or not
          * @param  {boolean} _x     Should reverse the color list
-         * @return { boolean | module} Is color list being reversed
+         * @return {boolean | module} Is color list being reversed
```
amber-eb: Could add "or chart module to chain calls" to keep in line with return documentation for most properties

src/charts/bar.js
```diff
@@ -977,7 +1002,7 @@ define(function(require) {
         /**
          * Changes the order of items given the custom function
          * @param  {Function} _x             A custom function that sets logic for ordering
-         * @return { (Function | Module) }   Updated module with ordering function applied
+         * @return {(Function | Module)}   Updated module with ordering function applied
```
amber-eb: Same here. Returns either the ordering function or the module with ordering function applied

src/charts/bar.js
```diff
@@ -793,12 +795,35 @@ define(function(require) {
...
+         * if you want to apply a custom behavior to the highlighted bar on hover.
+         * When hasSingleBarHighlight is true the highlighted bar will be the
+         * one that was hovered by the user. When hasSingleBarHighlight is false
+         * the highlighted bars are all the bars but the hovered one. The default
+         * highlight effect on a bar is darkening the highlighted bar(s) color.
+         * @param  {Function} _x        Desired operation operation on a hovered bar passed through callback
+         * @return {highlightBarFunction | module} Is highlightBarFunction used or Chart module to chain calls
+         * @public
+         * @example barChart.highlightBarFunction(e => d3.select(e).attr('fill', 'blue'))
+         * barChart.highlightBarFunction(e => null) // will disable the default highlight effect
```
amber-eb: Would it be possible to pass `null` here and instead check if the function exists before calling it? Seems odd to pass a function that returns `null`, but perhaps there's another reason this is necessary.

src/charts/bar.js
Golodhros: Go for it. 
We haven't been super strict with all that formating.

src/charts/bar.js
DalerAsrorov: Nice catch! Will fix those

src/charts/bar.js
```diff
@@ -105,6 +105,7 @@ define(function(require) {
             animationStepRatio = 70,
             interBarDelay = (d, i) => animationStepRatio * i,
 
+            highlightBarFunction = (bar) => d3Selection.select(bar).attr('fill', ({name}) => d3Color.color(colorMap(name)).darker()),
```
Golodhros: Now seeing this, I am wondering if it will make sense to pass the selection of the bar to the function:
```
(barSelection) => barSelection.attr('fill', ...
```
This could make it a bit easier for the user to pass the new function? Or not?
What do you guys think?

DalerAsrorov: @Golodhros I think this totally makes sense. In this case, we are abstracting the usage of d3 from the user - which is I think what we aim for? I don't see other cases where the bar rectangle `<rect>` tag would be used by `d3` other than select. Unless there is any more possible ways do something with the passed `bar`. In short, I agree.

src/charts/bar.js
dalerasrorov-eb: I just updated the code and added check for `null` or `undefined` or otherwise use an empty function. This way it's possible to do just `barChart.highlightBarFunction(null)`

Issue #461, opened 2017-12-27
Update Webpack Config using latests improvements from Britecharts-React


miglesiasEB:
## Expected Behavior

In Britecharts react we have added several improvements like the use of merge, the dashboard and more. Check https://github.com/eventbrite/britecharts-react/blob/master/webpack.config.js.

## Current Behavior

Our webpack file is not great. Too long and convoluted.

## Possible Solution

Make both webpack configs similar.

PR #518, opened 2018-02-20
Renaming Wepback config objects and using const/let

Golodhros:
## Description

Moves the webpack config file into ES6, renaming some objects to be more consistent with Britecharts-React config.

## Motivation and Context

First step on https://github.com/eventbrite/britecharts/issues/461

## How Has This Been Tested?

Ran tests and documentation

## Types of changes

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,21 +11,21 @@ const isProduction = env === 'prod' || env === 'prodUMD';
 const projectName = 'britecharts';
 
 const CHARTS = {
-    'bar': './src/charts/bar.js',
-    'donut': './src/charts/donut.js',
-    'legend': './src/charts/legend.js',
-    'line': './src/charts/line.js',
-    'tooltip': './src/charts/tooltip.js',
-    'miniTooltip': './src/charts/mini-tooltip.js',
-    'sparkline': './src/charts/sparkline.js',
-    'stackedArea': './src/charts/stacked-area.js',
-    'stackedBar': './src/charts/stacked-bar.js',
-    'groupedBar': './src/charts/grouped-bar.js',
-    'step': './src/charts/step.js',
-    'brush': './src/charts/brush.js',
-    'loading': ['./src/charts/helpers/load.js'],
+    bar: './src/charts/bar.js',
+    donut: './src/charts/donut.js',
+    legend: './src/charts/legend.js',
+    line: './src/charts/line.js',
+    tooltip: './src/charts/tooltip.js',
+    miniTooltip: './src/charts/mini-tooltip.js',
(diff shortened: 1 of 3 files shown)


dalerasrorov-eb:
LGTM

webpack.config.js
```diff
@@ -1,86 +1,88 @@
...
+    enforce: 'pre',
+    loader: 'eslint-loader',
+    options: {
+        emitWarning: true,
+        failOnError: false,
+    }
+};
 
-    plugins = [
+let plugins = [
```
amber-eb: Why `let` for this one?

webpack.config.js
```diff
@@ -1,86 +1,88 @@
...
+// Move to Parts
+const babelLoader = {
+    test: /\.js$/,
+    exclude: /(node_modules)/,
+    use: [{
+        loader: 'babel-loader',
         options: {
-            emitWarning: true,
-            failOnError: false,
+            presets: ['es2015', 'stage-0'],
```
amber-eb: What does the`stage-0`preset do?

webpack.config.js
```diff
@@ -1,86 +1,88 @@
...
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+const merge = require('webpack-merge');
+
+const env = require('yargs').argv.mode;
+const isProduction = env === 'prod' || env === 'prodUMD';
+
+const projectName = 'britecharts';
+
+const CHARTS = {
+    'bar': './src/charts/bar.js',
```
amber-eb: Is it necessary for the keys to be strings? Would it be possible to do:
```
const CHARTS = {
    bar: './src/charts/bar.js',
```

miglesiasEB: Yeah, totally

webpack.config.js
miglesiasEB: it's a Babel configuration: https://babeljs.io/docs/plugins/preset-stage-0/

webpack.config.js
miglesiasEB: In line 91, if the environment is production, we push something in.

That said, a const won't break if we do that, but it feels weird.

webpack.config.js
amber-eb: Just to clarify, what I am asking is whether it is necessary for the keys to be in quotes.

webpack.config.js
amber-eb: I see, so this conveys intent to change the values inside of the array.

dalerasrorov-eb: I think it is a good convention to declare variables with `let` if they are being mutated in the future (unless they are 3rd party modules or a custom module of some sort.)

PR #520, opened 2018-02-21
Updating Code styleguide with structure and helpers info

Golodhros:
## Description

Tunning the structure of our docs

## Motivation and Context

Fixes https://github.com/eventbrite/britecharts/issues/264


dalerasrorov-eb:
Do you think there should be a redirect from contributing to style guide? If a contributor looks at the `CONTRIBUTING` file, seems helpful to let the potential contributor to know about the helpers and the styleguide in some way

miglesiasEB:
Makes sense! Adding a link back!

PR #521, opened 2018-02-22
Added contributors and tags

DalerAsrorov:
Added contributors same way as in `britecharts-react`

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,18 @@
     {
       "name": "Sun Dai",
       "email": "sundaisf@gmail.com"
+    },
+    {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -45,12 +57,12 @@
...
-    "graph",
-    "reusable",
+    "javascript",
     "es6",
-    "es2015",
-    "visualization"
+    "visualizations",
+    "data-visualization",
+    "interactive-visualizations"
+
```
Golodhros: Let's add "svg" too for Britecharts

Issue #456, opened 2017-12-25
Add touchmove event to line and stacked area charts


Dieber:
Is it possible to consider adding a 'touchmove' event in order that updating position of tooltip when finger moving on linecharts?

Thanks n Merry X-mas

Golodhros:
Hi @Dieber , thanks for the proposal.
That makes sense and it shouldn't be hard to add a 'touchmove' event along with the current 'mousemove' events we are using.

dalerasrorov-eb:
@Dieber I am trying to understand what you mean by ```...in order that updating position of tooltip when finger moving on linecharts?```

Could you provide more details on what you need and what in this case `touchmove` mean? The `touchmove` event sounds more like a drag-and-drop movement, but it seems like it's not the same in your case. Would appreciate more information.

ISMAELMARTINEZ:
I believe @Dieber is talking about adding touch events to the charts (We are also really interested in that feature!!).

https://github.com/d3/d3-selection/blob/master/README.md#touch

It does seem to be possible to add it via d3 in a similar way as the mouse events are handle, but might be better creating a common "eventsHandling" object that all the charts can use (so to avoid code duplication).

We don't have bandwidth at the moment to develop it, but I am happy to review or comment and test that code. 

Thanks,

dalerasrorov-eb:
Thanks for helpful info @ISMAELMARTINEZ that sounds like a nice feature to add to charts. First we would create it for one chart, see how it is used, and then create a sharable function that can be used by other charts. Will keep you posted.

ISMAELMARTINEZ:
are you looking also into adding click and touch events? Thanks.

ISMAELMARTINEZ:
I will add a comment in that commit. Great job.

PR #523, opened 2018-02-23
Feaeture - touchmove event dispatch handlers for Line and StackedArea

dalerasrorov-eb:
## Description
Added `touchmove` custom event handlers for both `Line` and `StackedArea` charts. It can be called and handled without including tooltip.

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/456

## How Has This Been Tested?
Added 1 test each for StackedArea and Line charts that check the `customTouchMove` event handler is dispatched when `touchmove` event is called.

`touchmove` for Stacked Area:
barchart_touch_move_stacked_area

`touchmove` for Line:
barchart_touch_move

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -213,7 +213,8 @@ define(function(require){
                 'customMouseOver',
                 'customMouseOut',
                 'customMouseMove',
-                'customDataEntryClick'
+                'customDataEntryClick',
+                'customTouchMove'
             );
 
         /**
@@ -246,6 +247,8 @@ define(function(require){
                     drawVerticalMarker();
                     addMouseEvents();
                 }
+
+                addTouchEvents();
             });
         }
 
@@ -281,6 +284,17 @@ define(function(require){
                 })
                 .on('mousemove',  function(d) {
                     handleMouseMove(this, d);
+                });
+        }
+
+        /**
+         * Adds events to the container group for the mobile devices
+         * Adding: touch
+         */
+        function addTouchEvents() {
+            svg
+                .on('touchmove', function(d) {
+                    handleTouchMove(this, d);
                 });
         }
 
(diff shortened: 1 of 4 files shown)


ISMAELMARTINEZ:
Hi @DalerAsrorov, I saw your commit and the change is small so it should just be ok. 

I see there is a bit of inconsistency between the charts functionality mainly due code duplication (or missing between charts). 

I can see that the line chart does handle the "customDataEntryClick" that I assume is the click event, but other charts like the groupbar doesn't implement that part.

Are you planning to share the common functionality between charts? 

Another question is regarding the touchmove event, should it not be also implement the same functionality as the handleMouseMove function (a part from the dispacher)?

Hope helps,

Golodhros:
Hi @ISMAELMARTINEZ, 
'customDataEntryClick' is not a regular click event, is a click in one of the highlight points shown when hovering over the line or stacked area charts. The Bar Chart is the only one that has a 'customClick' event.

I am curious about your use case for having a click on the grouped bar. Still, I think it makes sense to add customClicks to all bar charts.

dalerasrorov-eb:
Will investigate more into adding the `touchmove` to simulate `mousemove`, seems like the API of adding event handlers changed multiple times.

ISMAELMARTINEZ:
@Golodhros, We show energy data for hours/days/weeks/months...  and use the group bar for comparison. 

when you click on the bar you "zoom in" to that part of the day/week/month. That is our usage.

Golodhros:
Thanks @ISMAELMARTINEZ !

I have created an issue for this: https://github.com/eventbrite/britecharts/issues/534

dalerasrorov-eb:
@ISMAELMARTINEZ I have a PR ready https://github.com/eventbrite/britecharts/pull/535, if you have comments feel free to add!

Issue #100, opened 2017-02-07
Create Sparkline variation based on Tufte sparkline


Golodhros:
We want to have a sparkline that works along with text.
The inspiration for this comes from Tufte's Beautiful Evidence book.

It should have a size related to font sizes, with a probable 'tunning' factor in order to support different font faces.

Golodhros:
![trump_budget_cuts_2019__what_trump_proposed_cutting_in_his_budget_-_washington_post](https://user-images.githubusercontent.com/190833/36521068-a36c1694-1749-11e8-86ca-0c60662519d9.png)

I liked this example from The Washington Post

DalerAsrorov:
This looks cool. I assume this is a still a d3 type `text` element that is tuned by setting font and size?

Golodhros:
Well, not really. It is still the same chart, but as it should work well integrated into text, it would probably benefit from a sizing relative to font size.
That said, this could be challenging as different fonts look different in size for the same font size :(

mfdj:
nitpick: `s/Tuftle/Tufte`

PR #531, opened 2018-02-25
Feature - ability to add title to Sparkline chart with custom font settings

dalerasrorov-eb:
After looking at the shared image in #100, it made sense to me to give Sparkline chart ability to have its own title with custom font settings supplied by the user (the default for each font attribute is applied if not given to the object). 

## Description
Proposed feature to allow users to add a "title" text at the top (with possible align settings in the future). Added two new methods to Sparkline's API:
* `titleText` - sets the string of the title text
* `titleTextStyle` - sets the text font style settings like `font-weight`, `fill`, `font-family`, and `font-style` (more may be added).
Also, added a new `text-group` to the container set which appends the title text if the `titleText` is set.

## Motivation and Context
Partially due to https://github.com/eventbrite/britecharts/issues/100

## How Has This Been Tested?
Added tests to `sparkline.unit.spec.js` that tests the setters and getters of the new methods and tests their presence in the dom as well.

screen shot 2018-02-24 at 2 07 59 am

screen shot 2018-02-24 at 2 09 37 am

## Types of changes

## Checklist:

### DIFF
--- a/demos/demo-sparkline.js
+++ b/demos/demo-sparkline.js
@@ -31,8 +31,7 @@ function createSparklineChart() {
         .isAnimated(true)
         .duration(1000)
         .height(containerWidth / 4)
-        .width(containerWidth)
-        .titleText('Growth')
+        .width(containerWidth);
 
     container.datum(dataset.data).call(sparkline);
 }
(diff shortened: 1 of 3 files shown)


Golodhros:
This looks like a nice feature, but looking back to #100, I think maybe the feature wasn't clear.

The idea was to create a variant of the sparkline that could be mixed with text, like in the middle of a line. This is a pic from the book:

img_20180224_170825

That said, this looks good and we should add it!

Issue #487, opened 2018-01-17
Horizontal gridlines do not update on Bar Chart


chinaowl:
## Expected Behavior

The horizontal dashed gridlines on a bar chart should be spaced evenly, and/or it would be nice to specify how many of those lines to draw.

## Current Behavior

I don't really understand the current behavior. If you look at [this pen](https://codepen.io/chinaowl/pen/EoOjdB), only the first chart has nicely spaced gridlines. I varied the data for the other three and there are always three lines grouped at the top 3 ticks for some reason.

## Steps to Reproduce (for bugs)

https://codepen.io/chinaowl/pen/EoOjdB

## Your Environment

* Version used: 2.4.11
* Browser Name and version: Chrome 63

chinaowl:
I have a very similar issue here: https://codepen.io/chinaowl/pen/opQxBK

It seems like the horizontal gridlines do not update when the data changes. I'm hoping to avoid having to destroy the entire chart and initialize everything again.

Golodhros:
Hi @chinaowl, thanks for the issue!

You are right, this should get updated automatically. I did a sweep some time ago doing a change like this, I guess I missed this one. We should fix it.

thanks!

PR #537, opened 2018-02-28
Cleaning gridlines before rendering to fix gridline bug when reloading

Golodhros:
## Description

Adding cleaning logic to the grid lines in all bar charts

## Motivation and Context

Fixes https://github.com/eventbrite/britecharts/issues/487

## How Has This Been Tested?

Tested manually, ran tests

## Types of changes

### DIFF
--- a/src/charts/stacked-bar.js
+++ b/src/charts/stacked-bar.js
@@ -408,6 +408,10 @@ define(function(require){
         function drawGridLines() {
             let scale = isHorizontal ? xScale : yScale;
 
+            svg.select('.grid-lines-group')
+                .selectAll('line')
+                .remove();
+
             if (grid === 'horizontal' || grid === 'full') {
                 svg.select('.grid-lines-group')
                     .selectAll('line.horizontal-grid-line')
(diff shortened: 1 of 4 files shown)


src/charts/stacked-bar.js
```diff
@@ -408,6 +408,10 @@ define(function(require){
         function drawGridLines() {
             let scale = isHorizontal ? xScale : yScale;
 
+            svg.select('.grid-lines-group')
+                .selectAll('line')
```
dalerasrorov-eb: Could we turn this into a util/helper function? It will most likely be used in more charts that will be built in the future.

dalerasrorov-eb: Something like:

```function cleanGridlines() { //code }```

dalerasrorov-eb:
```function removeGridLines(svg) {
     svg.select('.grid-lines-group').selectAll('line').remove()
}

Golodhros: Yeah!
We probably should refactor all the grid line logic into one helper like the axis one.
I will create a ticket for it!

Golodhros:
Created https://github.com/eventbrite/britecharts/issues/538 regarding this

Issue #483, opened 2018-01-12
Set explicit API for empty states


miglesiasEB:
## Current Behavior
Right now, we are using an empty data array within the stacked area chart in order to show an empty state (that is a line in the bottom of the chart). We have something similar on the donut chart.

## Expected
We should, for the sake of clarity, add 'shouldShowEmptyState' options to these charts. This explicit API will make more clear to our users what would happen, without having to depend on remembering which different shapes of the data generates which states.

## Context
Making our API easier to use and understand

Golodhros:
![](https://github.trello.services/images/mini-trello-icon.png) [Set explicit API for empty states #483](https://trello.com/c/Kl90FXAg/67-set-explicit-api-for-empty-states-483)

PR #539, opened 2018-02-28
RC-3.0 Feature - method for manually setting empty states

dalerasrorov-eb:
Added `shouldShowEmptyState` method to `donut` and `stacked-area` APIs to set the empty state manually (even if data is given) to see the empty states as described in https://github.com/eventbrite/britecharts/issues/483

## Description
Empty state will be known using `shouldShowEmptyState`. For the donut chart, it will show an empty slice with no info for the centered text (we can switch to make it show 0% if needed).
*Update:* changed logic in `drawLegend` to show `0%` if `shouldShowEmptyState` is `true`

## Motivation and Context
https://github.com/eventbrite/britecharts/issues/483

## How Has This Been Tested?
Added two new setter/getter, one for each chart.

`donut.sholdShowEmptyState(true)`:
screen shot 2018-02-28 at 3 40 20 pm

`stackedArea.sholdShowEmptyState(true)`:
screen shot 2018-02-27 at 5 16 44 pm

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -601,7 +601,7 @@ define(function(require) {
 
         /**
          * Gets or Sets the emptySliceColor of the chart. 
-         * If data is mepty or shouldShowEmptyState is set to true, the chart will render 
+         * If data is empty or shouldShowEmptyState is set to true, the chart will render 
          * an empty slice with a given color (light gray by default)
          * @param  {Object} _x emptyDataConfig object to get/set
          * @return {Object | module} Current config for when chart data is an empty array
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -319,7 +319,7 @@ define(function(require) {
          * @private
          */
         function drawLegend(obj) {
-            if (obj.data) {
+            if (obj.data && !shouldShowEmptyState) {
 
                 svg.select('.donut-text')
                     .text(() => centeredTextFunction(obj.data))
--- a/src/charts/donut.js
+++ b/src/charts/donut.js
@@ -822,7 +822,6 @@ define(function(require) {
 
         /**
          * Gets or Sets the shouldShowEmptyState state of the chart
-         * The default value is -60
          * @param  {Boolean} _x Whether or not to set chart to empty state
          * @return {Boolean | module} Current shouldShowEmptyState or Chart module to chain calls
          * @public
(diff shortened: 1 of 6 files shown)


src/charts/donut.js
```diff
@@ -156,7 +157,7 @@ define(function(require) {
                 if (highlightedSliceId) {
                     initHighlightSlice();
                 }
-                if (isEmpty && emptyDataConfig.showEmptySlice) {
+                if ((isEmpty && emptyDataConfig.showEmptySlice) || shouldShowEmptyState) {
```
Golodhros: Actually, I wanted to remove the 'isEmpty' variable altogether.
Also, I think we should get rid of 'emptyDataConfig', keeping only 'emptySliceColor' as a regular accessor.

src/charts/donut.js
```diff
@@ -819,6 +820,23 @@ define(function(require) {
             return this;
         };
 
+        /**
+         * Gets or Sets the shouldShowEmptyState state of the chart
+         * The default value is -60
```
Golodhros: Oops

src/charts/stacked-area.js
```diff
@@ -521,7 +523,10 @@ define(function(require){
          * @private
          */
         function cleanData(originalData) {
-            originalData = originalData.length === 0 ? createFakeData() : originalData;
+            originalData = 
+                originalData.length === 0 || shouldShowEmptyState
```
Golodhros: Let´s get rid of "originalData.length" here, so only if shouldShowEmptyState is true we will show the empty state.

src/charts/stacked-area.js
```diff
@@ -1343,6 +1348,23 @@ define(function(require){
          */
         exports.axisTimeCombinations = axisTimeCombinations;
 
+        /**
+         * Gets or Sets the shouldShowEmptyState state of the chart
+         * The default value is -60
```
Golodhros: Oops

src/charts/donut.js
DalerAsrorov: Copy-pasting is hard :)

miglesiasEB:
Merged into https://github.com/eventbrite/britecharts/tree/release/3.0.0-rc1

Thanks a lot!

PR #540, opened 2018-02-28
Feature - Added support numberFormat in tooltip

violetlight:
## Description
I added a getter/setter method to the Tooltip component, updated all references from `valueFormat` to `numberFormat`, and removed the old code for `valueFormat`'s references and getter/setter method.

## How Has This Been Tested?
I have added a test to ensure that the `numberFormat()` getter/setter is provided and updated previous references to `valueFormat` so that they now use `numberFormat`.

## Types of changes

## Checklist:

### DIFF
--- a/demos/demo-grouped-bar.js
+++ b/demos/demo-grouped-bar.js
@@ -54,6 +54,7 @@ function creategroupedBarChartWithTooltip(optionalColorSchema) {
             .topicLabel('values')
             .dateLabel('key')
             .nameLabel('stack')
+            .numberFormat('b')
             .title('Testing tooltip');
 
         // Note that if the viewport width is less than the tooltipThreshold value,
--- a/demos/demo-grouped-bar.js
+++ b/demos/demo-grouped-bar.js
@@ -54,7 +54,6 @@ function creategroupedBarChartWithTooltip(optionalColorSchema) {
             .topicLabel('values')
             .dateLabel('key')
             .nameLabel('stack')
-            .numberFormat('b')
             .title('Testing tooltip');
 
         // Note that if the viewport width is less than the tooltipThreshold value,
--- a/demos/grouped-bar.html
+++ b/demos/grouped-bar.html
@@ -23,7 +23,6 @@ groupedBar
 container.datum(dataset.data).call(groupedBar);
 
 chartTooltip
-    .numberFormat('b') // format values to binary
     .title('Testing tooltip');
 tooltipContainer = d3.select('.metadata-group');
 tooltipContainer.datum([]).call(chartTooltip);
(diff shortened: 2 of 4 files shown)


demos/demo-grouped-bar.js
```diff
@@ -54,6 +54,7 @@ function creategroupedBarChartWithTooltip(optionalColorSchema) {
             .topicLabel('values')
             .dateLabel('key')
             .nameLabel('stack')
+            .numberFormat('b')
```
dalerasrorov-eb: We can remove it from the official demo

src/charts/tooltip.js
```diff
@@ -7,6 +7,8 @@ define(function(require){
     const d3Transition = require('d3-transition');
     const d3TimeFormat = require('d3-time-format');
 
+    const NUMBER_FORMAT = null;
```
dalerasrorov-eb: Instead of having a constant for it, I would directly assign `null` in line `116`

src/charts/tooltip.js
```diff
@@ -111,6 +113,7 @@ define(function(require){
             topicsOrder = [],
 
             // formats
+            numberFormat = NUMBER_FORMAT,
```
dalerasrorov-eb: can use directly `null` and remove the `NUMBER_FORMAT` constant

violetlight:
Associated issue: https://github.com/eventbrite/britecharts/issues/350

dalerasrorov-eb:
Thanks for doing it! Looks good, just a few minor changes to be made.

demos/demo-grouped-bar.js
dalerasrorov-eb: Don't think we need it in demo as well

demos/grouped-bar.html
```diff
@@ -23,6 +23,7 @@ <h3>The code</h3>
 container.datum(dataset.data).call(groupedBar);
 
 chartTooltip
+    .numberFormat('b') // format values to binary
```
dalerasrorov-eb: Don't think this in demo as well

violetlight:
Thanks for reviewing!

src/charts/tooltip.js
```diff
@@ -672,6 +673,20 @@ define(function(require){
             return this;
         };
 
+        /**
+         * Gets or Sets the number format for the value displayed on the tooltip
+         * @param  {string} [_x=".2f"] Desired number format
```
dalerasrorov-eb: Sorry just noticed it. the `[_x=".2f"]` will show that ".2f" is the default value while the default is `null`. currently. In `minitooltip` the default is indeed `.2f` but it's `null` in this case.

PR #542, opened 2018-03-04
Fixed data refresh bug

### LINKED ISSUE #541: Highlight data points failing on data refresh
Bug on data refresh

## Expected Behavior
When refreshing data in a line chart, for example:
```js
container.datum(data).call(lineChart);
container.datum(data).call(lineChart);
```
It is expected that the highlight circles appear properly and errors aren't thrown.

## Current Behavior
Data point highlight circles appear in the wrong place, and errors are thrown with "cannot read property 'getTotalLength' of undefined. see: https://codepen.io/bratter/pen/YeRXbv

## Possible Solution
After some digging it looks like the error comes from drawLines (about line 640 in line.js). The paths property is being saved to the enter selection, not to the enter + update selection (looks like it hasn't been updated since upgrade from d3v3 maybe).

Solution is therefore just to wrap `lines.enter()...` with lines.merge()`. Will submit a PR now. 

See pen: https://codepen.io/bratter/pen/YeRXbv

## Your Environment
* Version used: 2.5.1
* Browser Name and version: Chrome 63
* Operating System and version: Windows 10

bratter:
Bug on data refresh, see #541.

## Description
Wrapped setting of paths property in line chart with lines.merge() to ensure the property contains both the enter and the update selection.

## Motivation and Context
Resolves #541.

## How Has This Been Tested?
- Added new test to line.spec.js that will throw an error with out the fix
- re-ran tests

## Types of changes

## Checklist:

### DIFF
--- a/test/specs/line.spec.js
+++ b/test/specs/line.spec.js
@@ -231,6 +231,16 @@ define([
 
                         expect(actual).toEqual(expected);
                     });
+
+                    it('should not throw error on mousemove', function() {
+                        let container = containerFixture.selectAll('svg'),
+                            newDataset = buildDataSet('withOneSource');
+
+                        // Need to refresh the data twice to get failure before fix
+                        containerFixture.datum(newDataset).call(lineChart);
+                        containerFixture.datum(newDataset).call(lineChart);
+                        container.dispatch('mousemove');
+                    })
                 });
             });
 
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -638,16 +638,17 @@ define(function(require){
             lines = svg.select('.chart-group').selectAll('.line')
                 .data(dataByTopic, getTopic);
 
-            paths = lines.enter()
+            paths = lines.merge(lines.enter()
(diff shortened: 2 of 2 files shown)


test/specs/line.spec.js
```diff
@@ -231,6 +231,16 @@ define([
...
+
+                    it('should not throw error on mousemove', function() {
+                        let container = containerFixture.selectAll('svg'),
+                            newDataset = buildDataSet('withOneSource');
+
+                        // Need to refresh the data twice to get failure before fix
+                        containerFixture.datum(newDataset).call(lineChart);
+                        containerFixture.datum(newDataset).call(lineChart);
+                        container.dispatch('mousemove');
+                    })
```
Golodhros: Awesome! Thanks for the test!

DalerAsrorov:
Great job!! We should have real time examples to detect this kind had of bugs!

Golodhros:
Yeah, variable data demos would go a long way!

PR #563, opened 2018-03-22
Feature+Fix - opacity with grid line fixes

dalerasrorov-eb:
## Description
* Feature: Added ability to set the opacity of the circles in scatter plot
* Fix: Added fix for gridlines that was found in demo PR. The grid lines functions were doing the reverse thing (drawHorizontal was drawing vertical and same for the opposite function)

## Motivation and Context
Opacity is good and fixing gridlines is great.

## How Has This Been Tested?
Added new getter/setter test for new method. Gridlines tests exist already

screen shot 2018-03-22 at 1 46 38 pm

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/scatter-plot.js
+++ b/src/charts/scatter-plot.js
@@ -114,6 +114,7 @@ define(function(require) {
             right: 0
         },
 
+        circleOpacity,
         maxCircleArea = 10,
 
         colorSchema = colorHelper.colorSchemas.britecharts,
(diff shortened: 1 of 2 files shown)


src/charts/scatter-plot.js
```diff
@@ -114,6 +114,7 @@ define(function(require) {
             right: 0
         },
 
+        circleOpacity,
```
Golodhros: I guess you would need to add a default one of 1?

src/charts/scatter-plot.js
```diff
@@ -451,6 +454,21 @@ define(function(require) {
             return this;
         };
 
+        /**
+         * Gets or Sets the circles opacity value of the chart
+         * @param  {number} _x                 Desired opacity of circles of the chart
```
Golodhros: and note it here

Issue #459, opened 2017-12-25
Add option to draw permanent highlight points to line chart


Dieber:
We want to allow users to keep the dots highlighted on the line chart.

They should probably have a slightly different styling as when we show them when hovering.

![](https://user-images.githubusercontent.com/17017234/34381051-fc6d1598-eb40-11e7-996b-cf144e0da919.png)

Golodhros:
Hi @Dieber,
I am not sure to understand this issue. Could you give us a bit more context?

Dieber:
@Golodhros 
Thx for your reply, I'm amazing that you response all my question even my English sucks lol.


I mean add a hollow circular point on each data, like this
![feature](https://user-images.githubusercontent.com/17017234/34381051-fc6d1598-eb40-11e7-996b-cf144e0da919.png)

Thanks

Golodhros:
Ah, yeah, that makes sense. Adding it as a proposal.

dalerasrorov-eb:
You can use `shouldShowAllDataPoints ` in the next release

PR #565, opened 2018-03-25
Feature - ability to highlight data points with function in Line chart

DalerAsrorov:
Added ability to show all data points with new function `shouldShowAllDataPoints` in *Line* chart.

## Description
Added new Line chart API function `shouldShowAllDataPoints`. The data points highlight look exactly like the points rendered when user hovers over a specific data point range in Line.

## Motivation and Context

https://github.com/eventbrite/britecharts/issues/459

## How Has This Been Tested?
* Added setter/getter test
* Added DOM inspection test that checks if new container group for data points was created and that each data point circle have correct attributes.

screen shot 2018-03-25 at 4 59 21 am

screen shot 2018-03-25 at 4 59 27 am

## Types of changes

## Checklist:

### DIFF
--- a/demos/css/britecharts.css
+++ b/demos/css/britecharts.css
@@ -95,6 +95,9 @@
 .bar-chart .percentage-label {
   fill: #666A73; }
 
+.line-chart .data-point-marks {
+  fill: #ffffff; }
+
 .line-chart .topic .line {
   fill: none;
   stroke-width: 2;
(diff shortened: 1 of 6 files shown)


DalerAsrorov:
@Golodhros we draw them in their own separate group (hover circles are technically on the top) so when you hover, the original circles cover the new ones but it is noticeable only in the DOM inspector. On hover just the y axis line shows up with tooltip as before. Would you prefer to draw it some other way? We could add glow effect on hover for these. Let me know what you think!

Golodhros:
LGTM!!

Thanks a lot for this Daler!!

PR #572, opened 2018-03-30
Addressed design notes and added x and y axis formatter functions

dalerasrorov-eb:
More improvements for Scatter Plot

## Description
* Added 0.24 opacity, noted by Sun that it's used in Stacked Area and other charts.
* `circleOpacity` sets `fill-opacity` rather than `opacity`. The outline color should be solid -- this follows Sun's notes as well
* Added tests for data points
* added `xAxisFormat` and `yAxisFormat` chart API methods for formatting x and y axis values
* Used correct class for both axis labels
* Added DOM tests for the new API functions

## Motivation and Context
Improvements and essential API methods

## How Has This Been Tested?
Added multiple tests to `scatter-plot.spec.js`
* Added tests for data points
* Added DOM tests for the new API functions

* Circle opacity default is 0.24 and stroke (outline) color is solid:

screen shot 2018-03-30 at 12 47 00 am

screen shot 2018-03-30 at 12 47 04 am

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/scatter-plot.js
+++ b/src/charts/scatter-plot.js
@@ -103,9 +103,10 @@ define(function(require) {
         maskGridLines,
 
         xAxis,
+        xAxisFormat = '',
         xScale,
         yAxis,
-        yAxisFormat = null,
+        yAxisFormat = '',
         yScale,
         areaScale,
         colorScale,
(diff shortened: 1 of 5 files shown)


src/charts/scatter-plot.js
```diff
@@ -713,6 +717,21 @@ define(function(require) {
             return this;
         };
 
+        /**
+         * Exposes ability to set the format of x-axis values
+         * @param  {String} _x               Desired height for the chart
```
Golodhros: Is this working fine with a format of ' ' as default?

DalerAsrorov: Yes! I believe ‘’ let’s it figure out the format on its own. But I tried null or undefined and it was throwing an error. If there is some other better ways you know, we can do that instead!

Golodhros: Awesome, sounds good!

miglesiasEB:
LGTM

PR #577, opened 2018-04-02
Feature - Scatter Plot tooltip impl using voronoi's diagram.find

dalerasrorov-eb:
Implementation of the tooltip movement and rendering using voronoi's d3 `.find` utility.

## Description
Voronoi's comes with D3 API [here](https://github.com/d3/d3-voronoi). It returns the nearest site to point [x, y] as mentioned in documentation. Using this method, we can show the data in the tooltip based on mouse [x, y] location by passing mouse coordinates to the `.find` voronoi diagram method.

* Added glow effect on hover of the calculated highlight point.
* Tooltip shows the *y-axis* value of the data points.

## Motivation and Context
Tooltip for Scatter Plot

## How Has This Been Tested?
* Added tests for tooltip render. Will try to add more tests in the future

britecharts_scatter_voronoi_glow_1

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Britecharts
-> Britecharts is a client-side **reusable Charting Library** based on [D3.js v4][1] that offers easy and intuitive use of charts and components that can be composed together to create amazing visualizations.
+> Britecharts is a client-side **reusable Charting Library** based on [D3.js v5][1] that offers easy and intuitive use of charts and components that can be composed together to create amazing visualizations.
 
 [![Build Status](https://travis-ci.org/eventbrite/britecharts.svg?branch=master)](https://travis-ci.org/eventbrite/britecharts)
 [![npm version](https://badge.fury.io/js/britecharts.svg)](https://badge.fury.io/js/britecharts)
--- a/src/charts/scatter-plot.js
+++ b/src/charts/scatter-plot.js
@@ -259,10 +259,16 @@ define(function(require) {
          * @private
          */
         function buildVoronoi() {
+            let voronoiWidth = width - margin.left - margin.right;
+            let voronoiHeight = height - margin.bottom - margin.top;
+
             voronoi = d3Voronoi.voronoi()
                 .x((d) => xScale(d.x))
                 .y((d) => yScale(d.y))
-                .extent([[0, 0], [width, height]])(dataPoints);
+                .extent([
+                    [0, 0],
+                    [voronoiWidth, voronoiHeight]
+                ])(dataPoints);
         }
 
         /**
@@ -550,7 +556,7 @@ define(function(require) {
             return {
                 closestPoint: voronoi.find(mousePos[0], mousePos[1]),
                 mousePos
-            }
+            };
         }
 
         /**
--- a/src/charts/scatter-plot.js
+++ b/src/charts/scatter-plot.js
@@ -103,7 +103,6 @@ define(function(require) {
         baseLine,
         maskGridLines,
         voronoi,
-        voronoiGroup,
 
         xAxis,
         xAxisFormat = '',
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 # Britecharts
-> Britecharts is a client-side **reusable Charting Library** based on [D3.js v4][1] that offers easy and intuitive use of charts and components that can be composed together to create amazing visualizations.
+> Britecharts is a client-side **reusable Charting Library** based on [D3.js v5][1] that offers easy and intuitive use of charts and components that can be composed together to create amazing visualizations.
```
Golodhros: Thanks!

src/charts/scatter-plot.js
```diff
@@ -213,6 +254,107 @@ define(function(require) {
                 .append('g').classed('metadata-group', true);
         }
 
+        /**
+         * Draws the voronoi component in the chart.
+         * @private
+         */
+        function buildVoronoi() {
+            let voronoiWidth = width - margin.left - margin.right;
```
Golodhros: this would be chartWidth as calculated in line 181, right?

src/charts/scatter-plot.js
```diff
@@ -312,120 +456,38 @@ define(function(require) {
...
                 .data(yScale.ticks(yTicks))
                 .enter()
-                .append('line')
-                .attr('class', 'horizontal-grid-line')
-                .attr('x1', (xAxisPadding.left))
-                .attr('x2', chartWidth)
-                .attr('y1', (d) => yScale(d))
-                .attr('y2', (d) => yScale(d))
+                  .append('line')
+                  .attr('class', 'horizontal-grid-line')
```
Golodhros: From this line on, the indentation should be 4 spaces

src/charts/scatter-plot.js
```diff
@@ -312,120 +456,38 @@ define(function(require) {
...
-        function cleanData(originalData) {
-            return originalData.reduce((acc, d) => {
-                d.name = String(d[nameKey]);
-                d.x = d[xKey];
-                d.y = d[yKey];
-
-                return [...acc, d];
-            }, []);
+                  .append('line')
+                  .attr('class', 'vertical-grid-line')
```
Golodhros: Same here

src/charts/scatter-plot.js
```diff
@@ -480,6 +542,59 @@ define(function(require) {
...
+        /**
+         * Calculates and returns
+         * @return {Object}
+         * @param {*} svg
+         * @private
+         */
+        function getPointProps(svg) {
+            let mousePos = d3Selection.mouse(svg);
+            mousePos[0] -= margin.left;
+            mousePos[1] -= margin.top;
```
Golodhros: These feel a bit weird, usually we don't need to do this kind of adjustments.

src/charts/scatter-plot.js
```diff
@@ -480,6 +542,59 @@ define(function(require) {
...
+
+        /**
+         * Handler called on mousemove event
+         * @return {void}
+         * @private
+         */
+        function handleMouseMove(svg, d) {
+            let { mousePos, closestPoint } = getPointProps(svg);
+
+            tooltip.update(getPointData(closestPoint), mousePos, [chartWidth, chartHeight]);
```
Golodhros: Usually we use the Tooltip as an external chart, could we do the same here?

Golodhros: I guess we would pass the 'getPointData(closestPoint)' instead of 'd' down in the dispatcher call.

src/charts/scatter-plot.js
```diff
@@ -489,6 +593,59 @@ define(function(require) {
...
+        }
+
+
+        /**
+         * Draws the perpendecular lines and labels
+         * for the highlighted data point
+         * @return {void}
+         * @private
+         */
+        function highlightLabelsForDataPoint() {
```
Golodhros: You could wait until somebody asks for it. At the end of the day, with the tooltip there is already a way to see the values

PR #583, opened 2018-04-04
Fix - scatter plot crosshair lines should not display over tooltip

DalerAsrorov:
## Description
Added a fix for the Scatter Plot. The crosshair lines showed up over the tooltip, so I placed them under path clip - in the same level as highlight point. I will add more tests in the future.

## Motivation and Context

The crosshair lines showed up on the top of tooltip, this should not happen.

## How Has This Been Tested?

Added a few more tests to make sure that highlight point shows up on hover

*Before:*
screen shot 2018-04-03 at 9 28 35 pm

*After:*
screen shot 2018-04-03 at 9 31 13 pm

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/scatter-plot.js
+++ b/src/charts/scatter-plot.js
@@ -611,7 +611,7 @@ define(function(require) {
                 .enter()
                   .append('line')
                     .attr('class', 'extended-x-line')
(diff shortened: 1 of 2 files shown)


dalerasrorov-eb:
LGTM :trollface:

PR #566, opened 2018-03-25
Improve Brush.on documentation

namoscato:
## Description

* Adds parameters and events.
* ~Moves documentation JavaScript source URLs to relevant paths.~

## How Has This Been Tested?

I generated the documentation locally via:

```
$ yarn run docs
```

## Screenshots (if appropriate):

screen shot 2018-03-25 at 3 17 42 pm

## Types of changes

- [x] Refactor (changes the way we code something without changing its functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the [code style guide](https://github.com/eventbrite/britecharts/blob/master/CODESTYLEGUIDE.md) of this project.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #585, opened 2018-04-05
Dispatch customBrushEnd event when clearing selection

### LINKED ISSUE #576: Brush Chart: customBrushEnd event not dispatched when clearing selection
## Expected Behavior

When clearing the selection of the Brush Chart (i.e. clicking on the chart without dragging), the `customBrushEnd` event should be dispatched, enabling external bindings to react to the change.

Related to this issue but a lower priority from my perspective: Is there a reason why the events dispatched by this chart do not match the [`d3-brush` event types](https://github.com/d3/d3-brush#brush_on) (`start`, `brush`, and `end`)? The current `customBrushStart` event seems to more closely resemble `brush` but is similarly not called on _start_ (i.e. mousedown).

## Current Behavior

Currently, `customBrushEnd` is not dispatched when the selection is empty:

https://github.com/eventbrite/britecharts/blob/84d1265a8163aab8e8f9b975e77c7bb1ebd1b650/src/charts/brush.js#L373

## Possible Solution

I think it would make sense to pass:

```javascript
[null, null]
```

as an argument to registered `customBrushEnd ` handlers, but I am open to suggestions.

This would be a minor breaking change, so I am not sure how that is handled.

1. Click on brush chart without dragging
2. `customBrushEnd` event is _not_ dispatched

namoscato:
## Description

* Resolves #576
* Removes unused code
* Removes events from API navigation dropdown menu

## How Has This Been Tested?

I updated the Brush Chart demo to illustrate these changes. Unfortunately, adding unit tests for these event listeners was going to be a lot easier said than done.

apr-04-2018 21-05-53

## Types of changes

## Checklist:

### DIFF
--- a/demos/src/demo-brush.js
+++ b/demos/src/demo-brush.js
@@ -11,11 +11,13 @@ require('./helpers/resizeHelper');
 
 
 function createBrushChart() {
-    let brushChart = brush(),
-        testDataSet = new dataBuilder.BrushDataBuilder(),
-        brushContainer = d3Selection.select('.js-brush-chart-container'),
-        containerWidth = brushContainer.node() ? brushContainer.node().getBoundingClientRect().width : false,
-        dataset;
+    const brushChart = brush();
+    const testDataSet = new dataBuilder.BrushDataBuilder();
+    const brushContainer = d3Selection.select('.js-brush-chart-container');
+    const containerWidth = brushContainer.node() ? brushContainer.node().getBoundingClientRect().width : false;
+    let dataset;
+
+    let elementDateRange = d3Selection.select('.js-date-range');
 
     if (containerWidth) {
         dataset = testDataSet.withSimpleData().build();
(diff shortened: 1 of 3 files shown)


src/doc/template/publish.js
```diff
@@ -406,18 +406,6 @@ function buildNav(members) {
...
-
-    members.events.forEach(function(e) {
-      if (!hasOwnProp.call(seen, e.longname)) {
-
-        nav.event.members.push(linkto(e.longname, e.longname.replace("module:", "")));
-      }
-      seen[e.longname] = true;
-    });
-
-  }
```
namoscato: I'm not sure if this is the best way to handle this, but this removes the Brush events that were inadvertently added to the API navigation dropdown menu in #566:

screen shot 2018-04-04 at 9 12 53 pm

demos/src/demo-brush.js
```diff
@@ -29,16 +31,20 @@ function createBrushChart() {
...
                 d3Selection.select('.js-end-date').text(format(brushExtent[1]));
 
-                d3Selection.select('.js-date-range').classed('is-hidden', false);
+                elementDateRange.classed('is-hidden', false);
             })
             .on('customBrushEnd', function(brushExtent) {
                 // eslint-disable-next-line no-console
                 console.log('rounded extent', brushExtent);
+
+                if (null === brushExtent[0]) {
```
dalerasrorov-eb: Is there a reason why we can't do:
```if (!brushExtent[0]) { ... }```

src/doc/template/publish.js
dalerasrorov-eb: To be honest, I have no idea how this templating works either. But we do not need those. Could you attach a screenshot of this and other changes? how does demo look like now?

demos/src/demo-brush.js
namoscato: No, we could do that. I was just being explicit; this should only ever be `null` or a `Date`.

src/doc/template/publish.js
namoscato: Here is what the API navigation dropdown menu looks like now:

screen shot 2018-04-04 at 10 11 56 pm

The Brush Chart demo functionality change is illustrated in the gif above.

demos/src/demo-brush.js
Golodhros: I would just reverse this to read:
   if (brushExtent[0] === null) ...

src/doc/template/publish.js
Golodhros: Ha, I don't have any idea either, this was tunned by another contributor.

Issue #594, opened 2018-04-09
No way to provide an alternative formatter for numbers


adrm:
## Expected Behavior

I would like to have some way to use any function to format numbers in tooltips or mini-tooltips. It would suffice to be able to set the valueFormatter property. I have not found a way to do this. Is it? Would you accept a pull request making this property available to the user of the library?

## Current Behavior

No way to format numbers without using d3-format.

## Possible Solution

Make valueFormatter accessible.

## Context

I am using this library in Angular app and I would like the Angular function (pipe) that I use to format every other number in the app to be used also in tooltips inside britecharts.

Golodhros:
Hi @adrm, thanks for the issue!

I see, we didn't think on that use case. I guess it makes sense and we could add it to those components.

If you are so kind to create a PR for both Tooltip and Mini-tooltip, that would be awesome!

adrm:
Perfect, I'm on it. I wanted to confirm with you that you are ok with this feature before implementing.

PR #598, opened 2018-04-10
valueFormatter setter and getter in tooltip and mini-tooltip

adrm:
## Description

Setter and getter for valueFormatter added to the tooltip and mini-tooltip charts.

## Motivation and Context

Fixes https://github.com/eventbrite/britecharts/issues/594

Without this patch, britecharts only allows you to format values with d3-format. In my case, I am using Angular and I would like to use the Angular number pipe to format every number, also those found inside britechart tooltips.

## How Has This Been Tested?

First, I tried to follow the contribution and style guides as close as possible, trying to mimic already existing code in case of doubt. Then, I tried to touch the least amount of code possible, and do so in the most simple way. Lastly, I ran the test suite with my new tests for the added features, and everything passed with no problem.

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/mini-tooltip.js
+++ b/src/charts/mini-tooltip.js
@@ -385,6 +385,7 @@ define(function(require){
          * @param  {Function} _x Desired formatter function
          * @return {Function | module} Current valueFormatter or Chart module to chain calls
          * @public
+         * @example tooltipChart.valueFormatter(value => value.toString().length.toString())
          */
         exports.valueFormatter = function(_x) {
             if (!arguments.length) {
--- a/src/charts/mini-tooltip.js
+++ b/src/charts/mini-tooltip.js
@@ -361,6 +361,7 @@ define(function(require){
                 return nameLabel;
             }
             nameLabel = _x;
+
             return this;
         };
 
@@ -375,6 +376,7 @@ define(function(require){
                 return numberFormat;
             }
             numberFormat = _x;
+
             return this;
         }
 
@@ -392,6 +394,7 @@ define(function(require){
                 return valueFormatter;
             }
             valueFormatter = _x;
+
             return this;
         }
 
@@ -417,6 +420,7 @@ define(function(require){
                 return title;
             }
             title = _x;
+
             return this;
         };
 
@@ -444,6 +448,7 @@ define(function(require){
                 return valueLabel;
             }
             valueLabel = _x;
+
             return this;
         }
 
--- a/src/charts/mini-tooltip.js
+++ b/src/charts/mini-tooltip.js
@@ -379,6 +379,22 @@ define(function(require){
         }
 
         /**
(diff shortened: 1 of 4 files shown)


src/charts/mini-tooltip.js
```diff
@@ -378,6 +378,22 @@ define(function(require){
...
         }
 
+        /**
+         * Gets or Sets the formatter function for the value displayed on the tooltip.
+         * Setting this property makes the tooltip ignore numberFormat. Set by default to
+         * d3-format formatter with numberFormat.
+         * @param  {Function} _x Desired formatter function
+         * @return {Function | module} Current valueFormatter or Chart module to chain calls
+         * @public
+         */
```
Golodhros: Could you add an @example tag showing how would you use this accessor?

src/charts/mini-tooltip.js
```diff
@@ -378,6 +378,22 @@ define(function(require){
...
+        exports.valueFormatter = function(_x) {
+            if (!arguments.length) {
+                return valueFormatter;
+            }
+            valueFormatter = _x;
+            return this;
```
Golodhros: Nit: space before return

test/specs/mini-tooltip.spec.js
```diff
@@ -166,5 +166,23 @@ define(['jquery', 'd3', 'mini-tooltip'], function($, d3, tooltip) {
                 expect(newNameLabel).toBe(testNameLabel);
             });
         });
+
+        it('should allow overriding the value formatting function', () => {
+            let expected = '8',
+                actual;
+
+            tooltipChart.valueFormatter(value => value.toString().length.toString());
```
Golodhros: This line could work great for that example

src/charts/mini-tooltip.js
adrm: You mean to add an empty line before? I copied the structure of other getter/setter methods in the source and I haven't seen an empty line there. If you want me to add it please confirm.

src/charts/mini-tooltip.js
adrm: Example added for tooltip and mini-tooltip

test/specs/mini-tooltip.spec.js
adrm: I used the line for the `@example` annotation. Please tell me if I should add anything more.

src/charts/mini-tooltip.js
DalerAsrorov: Hey @adrm - the formatting in this file is wrong. Please do add an empty line before return. If you want you can also change the format in the same file to keep it consistent. Thanks a lot!

adrm: Thanks for the explanation. Will do!

adrm: fixed

Issue #608, opened 2018-05-18
Support for gradients on bar chart


ronvenna:
I am using the horizontal bar chart and want to fill the bars with a gradient. I see you have support for gradients on line chart but not bar chart. Do you have any advice on if this is achievable?

## Expected Behavior
Something like a `colorGradient` option would be nice 
```
barChart
    .colorGradient(['#39C7EA', '#4CDCBA']) // pass in a start color to end color
    .width(500)
    .height(300);
```

## Current Behavior
This would be an additional field compared to the current `colorSchema` field
```
barChart
    .colorSchema(['#39C7EA', '#4CDCBA'])
    .width(500)
    .height(300);
```

## Possible Solution
Listed my solution above in the Expected Behavior section

## Context
Mostly for styling not a big issue but would be a nice to have. Since you have this for linecharts it only makes sense to make this available to all charts to un-fragment this feature. 

## Your Environment
* Version used: 2.8.6
* Browser Name and version: Chrome 66
* Operating System and version: Mac OSX 10.13
* Link to your project or codepen: Private project

dalerasrorov-eb:
Hi @ronvenna ! Thanks for creating the issue! This sounds interesting, I marked it as a proposal and will work on it some time to see if it's plausible (i.e. looks good to add it)

dalerasrorov-eb:
<img width="631" alt="screen shot 2018-05-18 at 1 05 36 am" src="https://user-images.githubusercontent.com/31934144/40216873-ae4e346c-5a37-11e8-8fb8-b53941da389b.png">
I started some work. Des this look like something you would want? It's orange to pink gradient

ronvenna:
@dalerasrorov-eb yes this looks great! I appreciate the fast response.

dalerasrorov-eb:
No problem @ronvenna ! This PR got merged: https://github.com/eventbrite/britecharts/pull/609 - should be available next release

PR #609, opened 2018-05-18
Feature - custom gradient color schema function for Bar chart

dalerasrorov-eb:
## Description
Added ability for user to specify the color schema of the Bar chart. It would be most useful in cases where there is a single category of data points, kind of like in Line chart. In this case however, I thought we would give user to decide when to use it.

*Need suggestions:*
Does the way gradient colors are rendered for horizontal and vertical bars look good? I am not as familiar with gradient ratios and color distribution. Would be great to get advice if it doesn't look good.

## Motivation and Context

https://github.com/eventbrite/britecharts/issues/608

## How Has This Been Tested?

Added multiple tests to `bar.spec.js` that check API and if the gradient colors are referenced is API function specifies the colors

screen shot 2018-05-18 at 12 59 55 am
screen shot 2018-05-18 at 1 31 23 am
screen shot 2018-05-18 at 1 09 41 am

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -15,6 +15,7 @@ define(function(require) {
     const {exportChart} = require('./helpers/export');
     const colorHelper = require('./helpers/color');
     const {bar} = require('./helpers/load');
+    const {uniqueId} = require('./helpers/number');
 
     const PERCENTAGE_FORMAT = '%';
     const NUMBER_FORMAT = ',f';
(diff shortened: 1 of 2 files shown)


src/charts/bar.js
```diff
@@ -293,6 +326,17 @@ define(function(require) {
...
         }
 
+        /**
+         * A utility function that checks if custom gradient
+         * color map should be applied if specified by the user
+         * @param {String} name - bar's data point name
+         * @return {void}
+         * @private
+         */
+        function chooseColor(name) {
```
miglesiasEB: Maybe I would call this 'pickColor' or 'computeColor'?

src/charts/bar.js
```diff
@@ -698,10 +742,25 @@ define(function(require) {
 
         // API
 
+        /**
+         * Gets or Sets the gradient colors of a bar in the chart
+         * @param  {String[]} _x Desired color gradient for the line (array of two hexadecimal numbers)
+         * @return {String[] | module} Current color gradient or Line Chart module to chain calls
+         * @public
+         */
+        exports.barGradient = function(_x) {
```
miglesiasEB: Maybe we could call this 'chartGradient', as the gradients are not being applied to each bar, but to the whole chart

PR #613, opened 2018-05-22
Fix - do not select tooltip text on frequent click, new class

dalerasrorov-eb:
## Description
The tooltip text selection happened on click. This should not happen.

## Motivation and Context

Should not happen.

## How Has This Been Tested?

Visual test

**Before:**
tooltip-select-before

**After:**
tooltip-select-after

tooltip-select-after-tooltip

## Types of changes

## Checklist:

### DIFF
--- a/src/styles/common/_tooltip.scss
+++ b/src/styles/common/_tooltip.scss
@@ -22,3 +22,11 @@ $tooltip-width: 1;
     border-radius: 2px;
 }
 
+.select-disable {
+    -webkit-touch-callout: none; /* iOS Safari */
+    -webkit-user-select: none; /* Safari */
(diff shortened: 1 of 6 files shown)


src/styles/common/_tooltip.scss
```diff
@@ -22,3 +22,11 @@ $tooltip-width: 1;
     border-radius: 2px;
 }
 
+.select-disable {
```
miglesiasEB: Maybe we could place this in britecharts/src/styles/base/_typography.scss as it will probably be used in other charts?

dalerasrorov-eb: Yes! I also added it to the big Tooltip (not just mini-tooltip) - see updated commit. Will add it to Typography.

PR #622, opened 2018-05-30
Fix - crosshair labels to be rendered on same level as lines

dalerasrorov-eb:
## Description

The way crosshair labels show up and disappear will be the same way as the crosshair lines. This should fix the bug when you hover over the crosshair labels that causes similar bug as lines had before.

## Motivation and Context

Bug

## How Has This Been Tested?

Added multiple DOM tests to `scatter-plot.spec.js`

Before:
scatter-plot-crosshair-labels-before

After:
scatter-plot-crosshair-labels-fix

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/scatter-plot.js
+++ b/src/charts/scatter-plot.js
@@ -573,14 +573,7 @@ define(function(require) {
          * @private
         */
         function drawDataPointsValueHighlights(data) {
-            removeDataPointsValueHighlights();
-
-            setCrossHairLinesStatus(1);
-
-            // initialize cross hair labels container
-            highlightCrossHairLabelsContainer = svg.select('.metadata-group')
-              .append('g')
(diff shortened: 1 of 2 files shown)


src/charts/scatter-plot.js
```diff
@@ -902,8 +896,9 @@ define(function(require) {
          * @param {Number}
          * @private
          */
-        function setCrossHairLinesStatus(status = 0) {
+        function setCrossHairComponentStatus(status = 0) {
```
miglesiasEB: A bit offtopic, but I would prefer a more explicit API for this, maybe showCorssHairComponent/hideCrossHairComponent?

dalerasrorov-eb: That makes sense! I will make more explicit functions in next PRs. Thanks for the suggestion!

Issue #299, opened 2017-06-29
Create Heatmap


miglesiasEB:
We love the heatmaps on github and Google Analytics, let's build our own!

## Expected Behavior
![analytics](https://user-images.githubusercontent.com/9063187/27715115-de42bb94-5ce9-11e7-994d-ea399554bcdf.png)

## Context
A heat map is a way to visualize data in tabular format, where in place of (or in addition to) the numbers, you leverage colored cells that convey the relative magnitude of the numbers.

We use color saturation to reduce the mental processing or reading the data, providing visual cues. It makes easy to get the min and max numbers easily.

It could show activity by hour and day or by day and month.

miglesiasEB:
Github's
![miglesiaseb__marcos_iglesias_](https://user-images.githubusercontent.com/9063187/27715081-a6d7aba6-5ce9-11e7-88a4-8011897777c9.png)

rwholey-eb:
@evb-sun !!!!!!!!!! 💃

evb-sun:
**Basic heatmap, more to come**
**[Invision](https://eventbrite.invisionapp.com/static-signed/live-embed/51160798/242147421/2/latest/0zfA613bclEAUbewnzmtnT38JwABiRBAf3A8I6sPBNlEw4izNMzBGHlEPyqNZaj3hYe1lEADHdlE5gJ4uuend7QbggQlE/Heatmap-Specs-2x.png)**

![heatmap specs 2x](https://user-images.githubusercontent.com/13459095/27845926-29c4035a-60e8-11e7-92d5-8a0d13a45b99.png)

evb-sun:
@miglesiasEB @rwholey-eb ^

miglesiasEB:
Great work @evb-sun! Let's do it!

gabitoju:
Hi!
Is there and estimate date for the release of this feature?

Golodhros:
Hi @gabitoju, thanks for asking!

This is actually at 30% more or less. I have a branch with an mvp, so it is highly likely that we are going to release this as soon as we get rid of the current bugs.

gabitoju:
If there's  anything even in pre-alpha that we could test that would be great!

Cospel:
Anyone has update on this? It looks to me as a very cool feature.

Golodhros:
This is going to be the next chart, hope I can get to it in this month.

dalerasrorov-eb:
Heatmap is implemented and available in `2.9.3` release.

PR #639, opened 2018-07-16
New Chart: initial stage of Heatmap chart

Golodhros:
## Description

A Heatmap chart based on the code from my blog post for Heart Internet.

tutorial__heatmap_chart

This is still very basic, in the next days I want to add:
* Vertical/Horizontal option
* Dynamic box sizes based on the width/height of the chart
* Date Label formats

## Motivation and Context
Heatmap chart was planned:
https://github.com/eventbrite/britecharts/issues/299

## How Has This Been Tested?
Added tests and demo

## Types of changes


src/charts/grouped-bar.js
```diff
@@ -36,9 +36,9 @@ define(function (require) {
      * @example
      * [
      *     {
-     *         "name": "2011-01",
-     *         "group": "Direct",
-     *         "value": 0
+     *         name: "2011-01",
+     *         group: "Direct",
+     *         value: 0
```
Golodhros: Making this consistent

src/charts/stacked-area.js
```diff
@@ -52,9 +52,9 @@ define(function(require){
      * @example
      * [
      *     {
-     *         "date": "2011-01-05T00:00:00Z",
-     *         "name": "Direct",
-     *         "value": 0
+     *         date: "2011-01-05T00:00:00Z",
+     *         name: "Direct",
+     *         value: 0
```
Golodhros: Same here

webpack.constants.js
```diff
@@ -2,38 +2,40 @@ const path = require('path');
...
+    'demo-heatmap': './demos/src/demo-heatmap.js',
     'demo-kitchen-sink': './demos/src/demo-kitchen-sink.js',
-    'demo-legend': './demos/src/demo-legend.js'
+    'demo-legend': './demos/src/demo-legend.js',
+    'demo-line': './demos/src/demo-line.js',
+    'demo-scatter-plot': './demos/src/demo-scatter-plot.js',
+    'demo-stacked-area': './demos/src/demo-stacked-area.js',
+    'demo-stacked-bar': './demos/src/demo-stacked-bar.js',
+    'demo-step': './demos/src/demo-step.js',
+    'demo-sparkline': './demos/src/demo-sparkline.js'
```
Golodhros: Ordering alphabetically

src/charts/heatmap.js
```diff
@@ -0,0 +1,350 @@
...
+            hoursHuman = [
+                '00h', '01h', '02h', '03h', '04h', '05h', '06h', '07h', '08h',
+                '09h', '10h', '11h', '12h', '13h', '14h', '15h', '16h', '17h',
+                '18h', '19h', '20h', '21h', '22h', '23h'
+            ],
+            hourLabelHeight = 20,
+
+            getValue = ({value}) => value,
+
+            toremove;
```
dalerasrorov-eb: what does this variable represent? Is this `toRemove`? maybe a different naming could help clear this out

src/charts/heatmap.js
```diff
@@ -0,0 +1,350 @@
...
+                .attr('width', boxSize)
+                .attr('height', boxSize)
+                .attr('x', ({hour}) => hour * boxSize)
+                .attr('y', ({day}) => day * boxSize)
+                .style('opacity', boxInitialOpacity)
+                .style('fill', boxInitialColor)
+                .style('stroke', boxBorderColor)
+                .style('stroke-width', boxBorderSize)
+                .transition()
+                    .duration(animationDuration)
```
dalerasrorov-eb: Is spacing here correct? Doesn't need two spaces instead of a tab?

src/charts/heatmap.js
```diff
@@ -0,0 +1,350 @@
...
+         */
+        function drawHourLabels() {
+            let hourLabelsGroup = svg.select('.hour-labels-group');
+
+            hourLabels = svg.select('.hour-labels-group').selectAll('.hour-label')
+                .data(hoursHuman);
+
+            hourLabels.enter()
+              .append('text')
+                .text((label) => label)
```
dalerasrorov-eb: I think it can be just `.text(label)`

src/charts/heatmap.js
```diff
@@ -0,0 +1,350 @@
...
+            let dayLabelsGroup = svg.select('.day-labels-group');
+
+            dayLabels = svg.select('.day-labels-group').selectAll('.day-label')
+                .data(daysHuman);
+
+            dayLabels.enter()
+              .append('text')
+                .text((label) => label)
+                .attr('x', 0)
+                .attr('y', (d, i) => i * boxSize)
```
dalerasrorov-eb: I see that there is multiple instances where `x * boxSize` where `x` is anything. Could be useful to create a private function `withBoxSize = (x) => x * boxSize` and then use like `.attr('y', (d, i) => withBoxSize(i))`. Doesn't have to be that way though, just a thought.

src/charts/heatmap.js
Golodhros: Oops! Great Catch!

src/charts/heatmap.js
Golodhros: Well, it seems we are not being consistent with this. I have seen no spaces, two spaces and a tab.

src/charts/heatmap.js
Golodhros: Hmmm... `label` is not a variable at this point.
This came from:
  .text((d) => d)

src/charts/heatmap.js
Golodhros: Yeah, that would make sense if we keep the functionality as it is right now.

Golodhros: However, I think that with the next steps, the boxes will stop being squares and start being rectangles.

Issue #636, opened 2018-07-14
Getting error for Stacked Bar chart: <rect> attribute y: Expected length, "NaN".


sagardalvi:
## Expected Behavior
Either error should be removed from the browser console or pass default 0 to the rectangle generated using d3 code

## Current Behavior
Getting too many errors in the browser console

## Possible Solution
This might be helpful :) 
https://stackoverflow.com/questions/40746319/error-rect-attribute-width-expected-length-nan-and-text-attribute-dx

## Steps to Reproduce (for bugs)
A height of generated SVG is 310
A width of generated SVG is 800

## Screenshots (if appropriate):
![image](https://user-images.githubusercontent.com/13640403/42722419-45e7069c-8769-11e8-9194-493a3e38b03e.png)
![image](https://user-images.githubusercontent.com/13640403/42722422-4fd26caa-8769-11e8-975b-edd52c2f23a6.png)

## Context

Getting too many errors in the browser console results in a bad impact on web page

## Your Environment
* Britechart Version used: 2.9.0
* Browser Name and version: Chrome
* Operating System and version: Mac OS X

Golodhros:
Hi @sagardalvi, thanks for creating this issue!

Could you create a codepen with the context of this bug? You could use this one https://codepen.io/Britecharts/pen/QqaogP as a starting point.

This way we will be able to help you better, thanks!!

sagardalvi:
Here is the example link
https://codepen.io/sagardalvi/pen/bjEZpK 

The primary observation regarding the issue is 
dataset created is as below
[
    {name: "A", value: 0.12, stack:"stack1"},
    {name: "A", value: 0.32, stack:"stack2"},
    {name: "B", value: 0.22, stack:"stack2"},
    {name: "C", value: 0.45,stack:"stack1"},
    {name: "D", value: 0.55,stack:"stack2"},
  ]

For name A we have 2 stacks [stack1 & stack2] But for B, C, D we have given only one stack (either stack1 or stack2) so while rendering it is giving errors on the browser console as shown in below screenshot.

![image](https://user-images.githubusercontent.com/13640403/42750071-8052c24e-8903-11e8-9db8-3ce854e0eae0.png)

dalerasrorov-eb:
Submitted fix, hope it resolves your issues. It will be available next release.

sagardalvi:
@dalerasrorov-eb Thank you for your support.

PR #640, opened 2018-07-16
Fix - Stacked Bar throwing console errors when columns have varied n of stacks

dalerasrorov-eb:
## Description

When each data point (A, B, C, D...) have different number of stacks (A: stack1, stack2, B: stack2, C: stack1). The layer `barJoin` sets the bar values by keys: 

```
            let stack3 = d3Shape.stack().keys(stacks),
                dataInitial = transformedData.map((item) => {
                        let ret = {};

                        stacks.forEach((key) => {
                            ret[key] = item[key];
                        });

                        return assign({}, item, ret);
                });

            layers  = stack3(dataInitial);
```

But not each data point has *2* stacks, as a result, it adds `NaN` value to `barJoin` layer:
screen shot 2018-07-16 at 3 30 20 am

This PR is a fix that checks for `isNaN` values before drawing data. 

## Motivation and Context

https://github.com/eventbrite/britecharts/issues/636

## How Has This Been Tested?

Ran tests

*Before:*
screen shot 2018-07-16 at 3 38 47 am

*After:*
No errors in console while the chart is the same:
screen shot 2018-07-16 at 3 32 29 am

## Types of changes

## Checklist:

### DIFF
--- a/demos/src/demo-stacked-bar.js
+++ b/demos/src/demo-stacked-bar.js
@@ -23,6 +23,16 @@ function createStackedBarChartWithTooltip(optionalColorSchema) {
         tooltipContainer,
         dataset;
 
+        let barData = {
+            data:[
+              {name: "A", value: 0.12, stack:"stack1"},
+              {name: "A", value: 0.32, stack:"stack2"},
+              {name: "B", value: 0.22, stack:"stack2"},
+              {name: "C", value: 0.45,stack:"stack1"},
+              {name: "D", value: 0.55,stack:"stack2"},
+            ]
+          };
+
     if (containerWidth) {
         dataset = testDataSet.with3Sources().build();
 
@@ -33,8 +43,8 @@ function createStackedBarChartWithTooltip(optionalColorSchema) {
             .grid('horizontal')
             .isAnimated(true)
             .stackLabel('stack')
-            .nameLabel('date')
-            .valueLabel('views')
+            .nameLabel('name')
+            .valueLabel('value')
             .betweenBarsPadding(0.3)
             .on('customMouseOver', function() {
                 chartTooltip.show();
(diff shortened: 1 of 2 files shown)


demos/src/demo-stacked-bar.js
```diff
@@ -132,15 +142,15 @@ function createHorizontalStackedBarChart(optionalColorSchema) {
...
+    // createHorizontalStackedBarChart();
 
     // For getting a responsive behavior on our chart,
     // we'll need to listen to the window resize event
     redrawCharts = function(){
         d3Selection.selectAll('.stacked-bar').remove();
 
         createStackedBarChartWithTooltip();
-        createHorizontalStackedBarChart();
+        // createHorizontalStackedBarChart();
```
Golodhros: Do you want to keep this code in the demo?

src/charts/stacked-bar.js
```diff
@@ -518,7 +519,13 @@ define(function(require){
 
             let barJoin = layerElements
                     .selectAll('.bar')
-                    .data((d) => d);
+                    .data((d) => {
+                        return d.map(item => {
+                            item[0] = getValOrDefaultToZero(item[0]);
+                            item[1] = getValOrDefaultToZero(item[1]);
```
Golodhros: We would always have 2 items here?

demos/src/demo-stacked-bar.js
dalerasrorov-eb: @Golodhros I will remove it! Just committed all the code by default

src/charts/stacked-bar.js
```diff
@@ -584,6 +585,21 @@ define(function(require){
...
 
+        /**
+         * Filter out unkown stacks/values in the bar layers
+         * @param {Object[]} d
+         * @return {Object[]} filteredData
+         * @private
+         */
+        function filterOutUnkownValues(d) {
+            return d.map(layerEls => {
+                for (let i = 0; i < layerEls.length; i++) {
```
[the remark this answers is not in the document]
dalerasrorov-eb:
```
            return d.map(layerEls => {
                layerEls = layerEls.map(dataLayer => {
                    dataLayer = getValOrDefaultToZero(dataLayer);
                    return dataLayer;
                });
                return layerEls;
            });
```
I tried something like that but the array contains object properties called `data` that needs to be preserved and it gets lost if I use `.map`.

PR #646, opened 2018-07-20
implemented excludeMutedEvents chart option

ConstantinoSchillebeeckx:
## Description
Added a chart option name `excludeMutedEvents` which will exclude the given events from being muted in cases where the tooltip is hidden because `tooltipThreshold` is reached.

## Motivation and Context
I'm working on an application where I still need access to the *onclick* event even when the tooltip is hidden (because width > `tooltipThreshold`. By providing a list of excluded events, the user can customize the behavior of exactly which events are muted.

## Screenshots:
image

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/stacked-bar.js
+++ b/src/charts/stacked-bar.js
@@ -114,6 +114,7 @@ define(function(require){
             hasReversedStacks = false,
 
             tooltipThreshold = 480,
+            excludeMutedEvents = [],
 
             yAxisLabel,
             yAxisLabelEl,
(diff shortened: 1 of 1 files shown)


dalerasrorov-eb:
Hi @ConstantinoSchillebeeckx , thanks for taking time to add this new feature. It would be better to create an issue first before making a PR.

ConstantinoSchillebeeckx:
Thanks for the helpful feedback; I'll keep that in mind for the future 👍

Issue #648, opened 2018-07-30
Export does not work in Internet Explorer


flacoman91:
## Expected Behavior
Clicking the Export button, or using the Export Chart should create an image and download the png of the chart.

## Current Behavior
An error is thrown. "Unable to get property 'replace' of undefined or null reference.

## Steps to Reproduce (for bugs)
1. Using Internet Explorer 11, go to any demo chart such as https://eventbrite.github.io/britecharts/tutorial-bar.html
2. Click on the Export button
3. image is not downloadable and the error appears in the console
4.

## Context
We still have people who use Internet Explorer 11 and Edge as their default browsers and they would not be able to download an image of charts.

## Your Environment
* Version used: 2.9.3
* Browser Name and version: Internet Explorer 11 (11.0.9600.19035)
* Operating System and version: Windows 7

Golodhros:
Hi @flacoman91, thanks for creating this issue!

Yes, we should fix this bug. I guess we haven't stated anywhere, but we are supposed to support IE11.

Thanks again!

dalerasrorov-eb:
Addressed in this PR: https://github.com/eventbrite/britecharts/pull/652

PR #652, opened 2018-08-02
IE 11 export chart fix

flacoman91:
## Description
this PR addresses:https://github.com/eventbrite/britecharts/issues/648
the changes provide a workaround and fixes for IE 11 not being able to handle SVG exports.

## Motivation and Context
Allows users to download SVGs

related links:
https://stackoverflow.com/questions/12592417/outerhtml-of-an-svg-element

I found that svg.outerHTML was failing, and I added a polyfill, but that would still throw a security error when you try to run the .toDataURL() method on an image.
https://stackoverflow.com/questions/20322745/ie-throws-security-error-when-calling-todataurl-on-canvas

I researched the issues and possible ways to fix the svg download issue in ie and found a few examples that recommended using fabric.js.

This was the easiest and cleanest solution that I found:
https://willkoehler.net/2014/11/07/client-side-solution-for-downloading-highcharts-charts-as-images.html

http://jsfiddle.net/willkoehler/vushe780/

## How Has This Been Tested?
I built the demos and tested the Export button to download a chart.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   "homepage": "http://eventbrite.github.io/britecharts/",
   "dependencies": {
     "base-64": "^0.1.0",
+    "canvg-browser": "^1.0.0",
     "d3": "^5.0.0",
     "lodash.assign": "^4.2.0"
   },
@@ -129,8 +130,8 @@
     "webpack": "^3.6.0",
     "webpack-bundle-analyzer": "^2.9.0",
     "webpack-dev-server": "^2.9.1",
-    "webpack-merge": "^4.1.1",
     "webpack-livereload-plugin": "^0.9.0",
+    "webpack-merge": "^4.1.1",
     "yargs": "^8.0.2"
   }
 }
--- a/src/charts/helpers/style.js
+++ b/src/charts/helpers/style.js
@@ -96,7 +96,7 @@ module.exports = (function() {
                 });
 
                 // fix for IE 11 not supporting svg.outerHTML
-                result = $('<div>').append($(elem).clone()).html();
+                result = new XMLSerializer().serializeToString(elem);
 
                 elements = [].map.call(elements, (el, i) => {
                     el.style.cssText = cssTexts[i];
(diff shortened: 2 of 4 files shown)


src/charts/helpers/style.js
```diff
@@ -95,7 +95,9 @@ module.exports = (function() {
                     }
                 });
 
-                result = elem.outerHTML;
+                // fix for IE 11 not supporting svg.outerHTML
+                result = $('<div>').append($(elem).clone()).html();
```
Golodhros: Mmmm... is this the jQuery '$'?

Golodhros: If so, we should probably use another method.

package.json
```diff
@@ -70,6 +70,7 @@
   "homepage": "http://eventbrite.github.io/britecharts/",
   "dependencies": {
     "base-64": "^0.1.0",
+    "canvg-browser": "^1.0.0",
```
Golodhros: Interesting, let me check out how large is this dependency.

src/charts/helpers/style.js
flacoman91: yes, sorry, I wanted to use something that wasn't jquery but this seemed like the shortest way of doing it.  I will look for another alternative, so at least it doesn't fail testing.

flacoman91: Ok, this XMLserializer solution works instead of jquery.
https://stackoverflow.com/a/45120072

miglesiasEB:
Hi @flacoman91 and sorry for the outrageous delay! :(

I am also sorry that we cannot merge this fix. "canvg-browser" adds a staggering 100Kb to all our chart bundles, and I am afraid the value that it adds doesn't pay off.

I still think this is a great fix, and it could make sense to someone which requirements highlight being able to export in IE11. I will keep a link to this one around and send it to new users looking for this fix, but we cannot merge it to the rest of the users.

Thanks a lot!

Issue #650, opened 2018-07-31
Tooltip issues in IE, Microsoft Edge


flacoman91:
## Expected Behavior
the information in tooltips in IE are sometimes hanging off of the tooltip. 

## Current Behavior
Depending on how you hover the mouse over a data point, the text sometimes goes off the tooltip.

## Steps to Reproduce (for bugs)
1. Use IE 11 or MS Edge
2. go to https://eventbrite.github.io/britecharts/tutorial-stacked-area.html or https://eventbrite.github.io/britecharts/tutorial-line.html
3. hover over tooltip

## Screenshots (if appropriate):
<img width="821" alt="screen shot 2018-07-31 at 4 09 31 pm" src="https://user-images.githubusercontent.com/1038306/43484869-9c47ddf0-94dd-11e8-9b41-fb66368bc7c4.png">
<img width="869" alt="screen shot 2018-07-31 at 4 07 11 pm" src="https://user-images.githubusercontent.com/1038306/43484878-9f43cd20-94dd-11e8-8837-041ec2811ff0.png">

## Your Environment

* Version used: 2.9.3
* Browser Name and version: Internet Explorer 11, Microsoft Edge
* Operating System and version: Windows 7, Windows 10

sagardalvi:
@Golodhros I am also facing the same issue for a tooltip when I have used a tooltip.valueFormatter() function. 
For some points, tooltip label value gets out of the tooltip boundary, If we slightly move the cursor then it re-renders correctly.

![image](https://user-images.githubusercontent.com/13640403/43707977-eb995a84-9986-11e8-9124-157522c0b8cc.png)

DalerAsrorov:
I don't have access to IE11 in Mac now but I will test it with IE11 tomorrow in a working machine. By looking at the code, there is a fishy line that may cause these issues:
https://github.com/eventbrite/britecharts/blob/master/src/charts/tooltip.js#L363

I tried to find a similar case where `tooltipRight.node().getBBox()` is called and and found a thread that suggests that it may not work well in both IE10 and IE11:
https://stackoverflow.com/questions/16833304/scalable-vector-graphicssvg-getbbox-not-working-in-internet-explorer-10

If I manually put 0 instead of `tooltipRight.node().getBBox().width` in that line mentioned, it shows up like this:
[...]

Issue #113, opened 2017-02-07
Tooltip rework


Golodhros:
We want to make the tooltip smart about the data he shows. 
Designs coming soon.

miglesiasEB:
![charts_wip_preview_mode_-_invision](https://cloud.githubusercontent.com/assets/9063187/22848176/615e1678-efa7-11e6-8f88-7852fac15901.png)

![charts_wip_preview_mode_-_invision](https://cloud.githubusercontent.com/assets/9063187/22848180/6dfc7384-efa7-11e6-9d03-800f410ec490.png)

Golodhros:
Let's do #134 first

Golodhros:
We should also improve the show animation on the tooltip, maybe with a fade-in fade-out technique.

hisnameisjimmy:
This would be an excellent solution, the mini-tooltip idea for single source line charts is exactly what I'm looking for. I'm gonna adjust the source files on the build in order to do this on my own, since I don't need multi-line charts.

Another thing, it would be really nice if some of the design could happen in CSS, ie being able to assign your own class to the tooltip and make adjustments through that. Exporting more adjustable parameters would also be nice, right now things like `tooltipWidth` are not useable through config.

Golodhros:
Thanks for the feedback @hisnameisjimmy!

I am curious about your request for a custom CSS class, could you use the standard classes on the tooltip?

Yeah, it will make sense to make the width configurable.

Regarding the output in a container div, you can do that right now. You could listen to the mouse events and do anything you want with that information.

Golodhros:
A bit of work done in https://github.com/britecharts/britecharts/pull/907

Golodhros:
One of the main issues of the Tooltip is that it doesn't listen for the chart height and width, and so it does not have any collision logic to avoid being cut.

Golodhros:
Done in 3.0.0 across #1049, #1050, #1051 and #1052: the tooltip measures the chart it is drawn in and never gets cut off (it flips and slides at the edges), it fades in and out and glides between positions, and `tooltip` and `miniTooltip` are one component with one contract for every chart. The migration guide's step 10 lists the one breaking change. Thanks for the write-up that started this.

PR #653, opened 2018-08-06
fix for IE 11 tooltip text going out of box

flacoman91:
## Description
In IE, it appears that it doesn't always get the rendered node width.  It looks like there is some sort of lag for the element to get drawn and width calculated.

## Motivation and Context
This fix saves the tooltipRight width and updates it if it is found. 
link to issue: https://github.com/eventbrite/britecharts/issues/650

## How Has This Been Tested?
I made the change and tested it in a virtualbox image running windows 7 and IE 11
the right side text does appear a bit off sometimes, but its better than having it run outside of the box.

The long text titles is still having issues and doesn't wrap all the time.  i think it's the same issue.
https://github.com/eventbrite/britecharts/blob/master/src/charts/tooltip.js#L530

sometimes the text appears left or right justified
screen shot 2018-08-06 at 2 22 47 pm
screen shot 2018-08-06 at 2 23 11 pm

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/tooltip.js
+++ b/src/charts/tooltip.js
@@ -82,7 +82,7 @@ define(function(require){
             tooltipBorderRadius = 3,
             ttTextX = 0,
             ttTextY = 37,
-            textSize,
+            textHeight,
             entryLineLimit = 3,
             initialTooltipTextXPosition = -25,
             tooltipTextLinePadding = 5,
@@ -322,7 +322,6 @@ define(function(require){
          */
         function updateTopicContent(topic){
             let name = topic[nameLabel],
-                textHeight,
                 tooltipRight,
                 tooltipLeftText,
                 tooltipRightText,
@@ -350,12 +349,10 @@ define(function(require){
                 .style('fill', tooltipTextColor)
                 .text(tooltipRightText);
 
-            textSize = elementText.node().getBBox();
-
             // IE11 give us sometimes a height of 0 when hovering on top of the vertical marker
             // This hack fixes it for some cases, but it doesn't work in multiline (they won't wrap)
             // Let's remove this once we stop supporting IE11
-            textHeight = textSize.height ? textSize.height : 18.4;
+            textHeight = elementText.node().getBBox().height ? elementText.node().getBBox().height : textHeight;
 
             tooltipHeight += textHeight + tooltipTextLinePadding;
             // update the width if it exists because IE renders the elements
--- a/src/charts/tooltip.js
+++ b/src/charts/tooltip.js
@@ -86,7 +86,7 @@ define(function(require){
             entryLineLimit = 3,
             initialTooltipTextXPosition = -25,
             tooltipTextLinePadding = 5,
-
+            tooltipRightWidth,
             // Animations
             mouseChaseDuration = 100,
             ease = d3Ease.easeQuadInOut,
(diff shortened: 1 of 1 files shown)


Golodhros:
Hi @flacoman91, thanks a lot for the fix!

It looks good to me, does it fix the overflowing text we saw in the first screenshot of the issue?

flacoman91:
Ok, here goes.  Using the referenced code in the stackoverflow link that @DalerAsrorov pointed to, i made an adjustment to get the size. 
instead of the long text showing up outside of the box, it looks like the 2nd pic below. I'll keep digging to see if there is something else I can do

screen shot 2018-08-06 at 4 38 11 pm

screen shot 2018-08-06 at 4 38 29 pm

DalerAsrorov:
That is a great effort @flacoman91 - it looks much better than before. Does it still look good in other browsers (Chrome, Firefox)?

flacoman91:
It looks good in the other browsers.

This would be a stopgap solution, and I think that the positioning should be precomputed in a similar manner to how charts are built with the Axis yscale/xscales are done in bar charts.

Our project hasn't gotten to a point yet where we are going to need more than one value in the tooltip nor do we have titles that are that long yet, otherwise I can devote more time to this.

Golodhros:
Thanks a lot @flacoman91!

Thanks for your work on this issue!

We are working on a [rework of the tooltips](https://github.com/eventbrite/britecharts/issues/113), and all these issues will come at hand when actually going over it.

At this moment we are using the canonical text wrapping method that Bostock shared in [this block](https://bl.ocks.org/mbostock/7555321). We know it doesn't work well in IE, but maybe we could use this other project https://github.com/vijithassar/d3-textwrap and see how it works.

src/charts/tooltip.js
```diff
@@ -527,7 +525,17 @@ define(function(require){
...
-                    if (tspan.node().getComputedTextLength() > width) {
+                    // fixes for IE wrap text issue
+                    let temp;
+                    temp = document.createElementNS("http://www.w3.org/2000/svg","text");
+                    temp.appendChild(document.createTextNode(line.join(' ')));
+                    temp.setAttribute('id', 'tempnode');
+                    temp.setAttribute("x",38);
+                    temp.setAttribute("y",18);
+                    document.getElementsByTagName("svg")[0].appendChild(temp);
+                    const textWidth = document.getElementsByTagName("svg")[0].lastChild.getBBox().width;
```
Golodhros: I understand we do this to measure the text, right?

We have a small function in our text helper, `getTextWidth`, would this function work?

flacoman91: Yes, its to measure the rendered width.  I think the issue is with IE not knowing the size of anything until it's in the DOM.  this link https://stackoverflow.com/a/25168951 shows how to get the correct size in internet explorer

Golodhros: I see, I wonder if our method would work in IE11, as it is based in canvas

flacoman91: I just changed that to 

```
const textWidth = getTextWidth(line.join(' '), 16, 'Karla, sans-serif');
```

and it does provide the exact same functionality.  You have to pass the font size and font family into the function and it'll work.

Golodhros: Awesome!
That function has a default font size and font family, but they are different. Maybe we should update those?

Happy that this worked!

flacoman91:
Ok, i made the change to use the text width helper. 

It still can look like the bunched up text occasionally, but at least this way we're keeping things DRY.  I think it's more performant too since we're not constantly adding and removing things from the DOM.

src/charts/tooltip.js
```diff
@@ -540,8 +543,11 @@ define(function(require){
                                 .text(word);
                         }
                     }
+
+                    // temp.parentNode.removeChild(temp);
```
Golodhros: Could we remove this commented code?

flacoman91: whoops, that slipped. thanks.

Golodhros:
Thanks a lot for this contribution @flacoman91 !!

PR #660, opened 2018-09-07
Changed handeMouseOut to use computeColor instead of colorMap

mhammo:
## Description

I've changed handleMouseOut in bar.js so it uses computeColor to determine the fill attribute instead of colorMap, to bring it in line with all the other .attr('fill', x) method calls in the .js file.

## Motivation and Context

https://jsbin.com/pawuxoruti/1/edit?html,js,output

When using a gradient on a bar chart, if you hover over the chart it'll remove the gradient.

## How Has This Been Tested?

I've ran all yarn tests, and I'm also using this patch in production.
It's a very minor change, very unlikely to break anything.

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -110,7 +110,14 @@ define(function(require) {
             animationStepRatio = 70,
             interBarDelay = (d, i) => animationStepRatio * i,
 
-            highlightBarFunction = (barSelection) => barSelection.attr('fill', ({name}) => d3Color.color(colorMap(name)).darker()),
+            highlightBarFunction = (barSelection) => 
+                barSelection.attr('fill', ({name}) => 
+                    d3Color.color(
+                        chartGradientColors 
+                        ? chartGradientColors[1]
+                        : colorMap(name)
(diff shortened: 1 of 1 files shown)


Golodhros:
Hi @mhammo, thanks for the fix!!

This looks good, could I ask you to change it too in the `highlightBarFunction` definition?

Thanks a lot!

mhammo:
Hi @Golodhros,

Thanks for taking a look!
I'm on holiday abroad for a week, I'll take a look at the highlight when I get back.

I'm guessing d3.color().darker() won't work with a gradient URL, if not I'll generate a darker gradient in the gradient function, and hook that into highlightBarFunction.

Golodhros:
Oh! Good call @mhammo!!

Maybe we could simply darken the last item in the gradient?

mhammo:
Sounds good! I've added a conditional operator to use the last gradient color if it exists.

Golodhros:
LGTM

Thanks a lot @mhammo!!

Issue #661, opened 2018-09-10
Create a Demo page for the Color schemas


miglesiasEB:
## Expected Behavior
We should provide a demo section to show the different color palettes we support.

## Current Behavior
At this moment, you will need to dive into the code to discover the different color palettes.

## Possible Solution
A demo page for colors?

Golodhros:
I just saw that lyft open sourced a color tool: https://github.com/lyft/coloralgorithm
You can try it in https://www.colorbox.io/

Wouldn't be great to have something like that applying it to a chart!
It could be part of the sandbox too!

sennab:
Fancy! That would be super cool. I'm working on a pretty simple solution though to start for the demo pages :)

miglesiasEB:
Perfect!

PR #668, opened 2018-10-04
Add a demo page for color palettes

snazbala:
## Description
This adds a simple demo page to display the color palettes we offer.

## Motivation and Context
There's no way currently to see the color palettes we offer or how to use a palette in your chart. 
Fixes [#661](https://github.com/eventbrite/britecharts/issues/661.)

## How Has This Been Tested?
tested in Chrome and Firefox. Uses flexbox so it will likely break in IE11, do we care about this for the docs page though?

screen shot 2018-10-03 at 6 37 21 pm

## Types of changes

None of the above -- just docs

## Checklist:

- [n/a] I have added tests to cover my changes.


demos/src/demo-color.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const d3Selection = require('d3-selection');
+const colors = require('./../../src/charts/helpers/color');
+
+function createColors() {
+    const { colorSchemas } = colors;
+
+    const colorPalettes = Object.keys(colorSchemas).map(colorPalette => {
```
DalerAsrorov: That's awesome that you use functional approach to iterate over both color palettes and colorSchemas! I would suggest to use `forEach` instead since we use `.map` usually to return a new array with the same length. Since we are not using the `colorPalletes` array anywhere (you inject elements into the DOM), I would suggest to use `forEach` in both cases instead. Makes it more clear as to how it's being used. Nothing will change, `map` returns something and `forEach` doesn't. That's the only difference.

demos/src/demo-color.js
```diff
@@ -0,0 +1,29 @@
...
+            const div = document.createElement('div');
+
+            div.style.width = '100px';
+            div.style.height = '100px';
+            div.style.background = color;
+            paletteDiv.appendChild(div);
+        });
+    });
+
+    return colorPalettes;
```
DalerAsrorov: If you use `forEach`, you can remove `colorPalettes` variable and not return anything.

demos/src/demo-color.js
snazbala: Good point! Will update

PR #669, opened 2018-10-05
Add an option to display border in stacked area

tgallice:


tgallice:
Sorry about that, didn't whent o open it now...

DalerAsrorov:
Hi @tgallice - before making PRs please create an issue so that we understand what the use case is and what you are trying to do. Otherwise it may be the case that PRs won't ever get merged. Thanks!

PR #678, opened 2018-10-18
Fix - tooltip render error in firefox

davegomez:
## Description
Instead of applying the styles display: none and display: block, we need to change visibility attribute instead. The visual behavior remains the same but the tooltip DOM node remains in the chart and hence error is not being thrown.

## Motivation and Context
[NS_UPDATE javascript error in Firefox](https://github.com/eventbrite/britecharts-react/issues/75)

## How Has This Been Tested?
Revised the tests to test for visibility style attribute instead of display

fix-firefox-tooltip

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the [code style guide](https://github.com/eventbrite/britecharts/blob/master/CODESTYLEGUIDE.md) of this project.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/src/charts/tooltip.js
+++ b/src/charts/tooltip.js
@@ -163,7 +163,7 @@ define(function(require){
                 svg = d3Selection.select(container)
                   .append('g')
                     .classed('britechart britechart-tooltip', true)
-                    .style('display', 'none');
+                    .style('visibility', 'hidden');
 
                 buildContainerGroups();
                 drawTooltip();
@@ -647,7 +647,7 @@ define(function(require){
          * @public
          */
         exports.hide = function() {
-            svg.style('display', 'none');
+            svg.style('visibility', 'hidden');
 
             return this;
         };
@@ -735,7 +735,7 @@ define(function(require){
          * @public
          */
         exports.show = function() {
-            svg.style('display', 'block');
+            svg.style('visibility', 'visible');
 
             return this;
         };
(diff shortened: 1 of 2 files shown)


Golodhros:
Hi @davegomez, thanks a lot for this PR!!

Could you check this out as well in the Mini-tooltip??

Thanks a lot again!!

davegomez:
Hi @Golodhros. Yes, the mini-tooltip is using the same approach to fix the issue.

davegomez:
@Golodhros do you have an idea of when this might be live? Our QA is getting nuts with the bug and doesn't want to release before is fixed.

I also wonder if will be necessary to update the dependencies in britecharts-react?

Golodhros:
We could release right now, let me do it.

Yes, we should update the dependencies in Britecharts-React, but I won't have time next week.

davegomez:
I appreciate it @Golodhros. I think I might be able to update the dependencies in `britecharts-react` and make a PR.

I'm also looking into a bug I experiencing in `britecharts-react` when using a link created with `react-router` that breaks the chart if I click in the same link for the page I already loaded.

The error is thrown [here](https://github.com/eventbrite/britecharts-react/blob/master/src/charts/tooltip/Tooltip.js#L147) because the reference `this._rootNode` is null.

Not sure if you know something about it.

Issue #670, opened 2018-10-08
Make stacked-area Outline optional


tgallice:
## Expected Behavior
Disable stacked-area outline

## Current Behavior
The stacked-area outline is by default enabled and can not be disabled / removed.

## Possible Solution
Make the outline rendering optional. We can add an option set up by a method like `disableOutline(x bool)` or `hideOutline(x bool)` to handle this.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context
An outline was added to the stacked-area which is nice, but this changed the rendering of the stacked-area and It's not possible to go back to the previous style. For my usage, I need to display the stacked-area without the outline

## Your Environment

* Version used:
* Browser Name and version:
* Operating System and version:
* Link to your project or codepen:

Golodhros:
Hi @tgallice, thanks for the issue!

About this issue, I would simply add a stylesheet rule to hide it:
```
.area-outline {
    display: none;
}
```

Adding a boolean as you suggested wouldn't be hard, but this way we avoid doing anything. Would this work for you?

Golodhros:
Well, it is true you will be drawing elements that you don't need, but unless they are a lot of them it should be fine.

In which way do you think is not flexible?

tgallice:
Sorry for the delay. An option seems much easier to me to use. No need to add a global/precise rule for each graph you have. All the behavior is handled by the config of the graph, insted of a css rule that overrides default ones.

Golodhros:
I see. The API could be something like:
```
stackArea.hasOutline(true/false)
```
With the default set as true.

PR #679, opened 2018-10-21
Feature - make Stacked Area outlines optional

DalerAsrorov:
## Description

Resolves: https://github.com/eventbrite/britecharts/issues/670

## Motivation and Context

https://github.com/eventbrite/britecharts/issues/670

## How Has This Been Tested?

Added multiple tests

screen shot 2018-10-21 at 3 23 34 pm

## Types of changes

## Checklist:

### DIFF
--- a/src/charts/stacked-area.js
+++ b/src/charts/stacked-area.js
@@ -156,6 +156,8 @@ define(function(require){
             ease = d3Ease.easeQuadInOut,
             areaAnimationDuration = 1000,
 
+            hasOutline = true,
+
             svg,
             chartWidth, chartHeight,
             data,
@@ -833,6 +835,12 @@ define(function(require){
                     .style('stroke', ({key}) => categoryColorMap[key]);
             }
 
+            if (!hasOutline) {
+                svg.select('.chart-group')
+                  .selectAll('.area-outline')
+                  .style('display', 'none');
+            }
+
             // Exit
             series.exit()
                 .transition()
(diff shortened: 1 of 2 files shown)


src/charts/stacked-area.js
```diff
@@ -833,6 +835,12 @@ define(function(require){
                     .style('stroke', ({key}) => categoryColorMap[key]);
             }
 
+            if (!hasOutline) {
```
Golodhros: Have you think about moving the conditional to the line 819?

src/charts/stacked-area.js
```diff
@@ -1197,6 +1205,15 @@ define(function(require){
             return this;
         };
 
+        exports.hasOutline = function(_x) {
+            if (!arguments.length) {
+                return hasOutline;
+            }
+            hasOutline = _x;
```
sennab: Could you explain what the `_x` parameter is?

src/charts/stacked-area.js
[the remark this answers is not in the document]
Golodhros: Hahaha, it seemed it worked, just not fast enough!

src/charts/stacked-area.js
Golodhros: _x is an optional parameter that if passed, will be set as the new ´hasOutline´value, if not, then this function will return the current ´hasOutline´ variable value.

This pattern is effectively a way of having a getter and a setter all at once.
_x could be called anything different for sure, this is the convention we adopted though.

sennab: Thanks for the explanation! Makes sense, I hadn't seen that convention before.

src/charts/stacked-area.js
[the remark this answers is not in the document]
DalerAsrorov: I tried moving those to line `819` but the outline still shows up a little bit in animated one.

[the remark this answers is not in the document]
Golodhros: OK, thanks for trying!

Issue #682, opened 2018-10-25
Stacked Area Chart circle markers has incorrect positions when >10 areas


flacoman91:
## Expected Behavior
The chart should render correctly
## Current Behavior
When supplying the Stacked Area chart with more than 10 areas, the hover over of each data point does not have the correct points, and it causes the circles to appear at the wrong position on the chart.

## Possible Solution
this sort algorithm needs to be fixed.
https://github.com/eventbrite/britecharts/blob/a68518e7543902c14e001705033dcd70711438cf/src/charts/stacked-area.js#L1024

PR to come. 

## Steps to Reproduce (for bugs)
1. sample codepen https://codepen.io/flacoman91/pen/gBBxBd
2. you can comment out one of the areas from the dataset and things will be fixed
3.
4.

## Screenshots (if appropriate):
Incorrect positions currently
![incorrectdot-positions](https://user-images.githubusercontent.com/1038306/47524109-baae0d00-d867-11e8-9eec-f697011497a1.gif)

It should look like this:
![correctdot-positions](https://user-images.githubusercontent.com/1038306/47524393-838c2b80-d868-11e8-9183-caf965aca543.gif)

## Context
Just adding more than 10 areas causes this bug and it does not look correct.

## Your Environment
* Version used: britecharts 2.9.7
* Browser Name and version: Chrome / Microsoft IE, Edge
* Operating System and version: Mac OS, Windows
* Link to your project or codepen: https://codepen.io/flacoman91/pen/gBBxBd

PR #683, opened 2018-10-25
Stacked area chart incorrect circle positions when > 10 areas

flacoman91:
## Description
fixing the sort algorithm to correctly transform the circle on hover over.

## Motivation and Context
Fixes this issue
https://github.com/eventbrite/britecharts/issues/682

## How Has This Been Tested?
Created sandbox and sample data and updated the code.

## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/src/charts/stacked-area.js
+++ b/src/charts/stacked-area.js
@@ -1019,12 +1019,9 @@ define(function(require){
             cleanDataPointHighlights();
 
             // ensure order stays constant
-            let sortedValues = [];
-            values = values.filter(v => !!v);
-
-            for(let i = 0; i< order.length; i++){
-                sortedValues.push(values.find(o=> o.name === order[i]));
-            }
+            let sortedValues = order.reduce((acc, current) => {
+                return [...acc, values.find(({name}) => name === current)];
+            },[]);
 
             sortedValues.forEach((d, index) => {
                 let marker = verticalMarkerContainer
--- a/src/charts/stacked-area.js
+++ b/src/charts/stacked-area.js
@@ -1019,11 +1019,14 @@ define(function(require){
             cleanDataPointHighlights();
 
             // ensure order stays constant
-            values = values
-                .filter(v => !!v)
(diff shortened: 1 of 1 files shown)


src/charts/stacked-area.js
```diff
@@ -1019,11 +1019,14 @@ define(function(require){
...
-                .filter(v => !!v)
-                .sort((a,b) => order.indexOf(a.name) > order.indexOf(b.name));
+            let sortedValues = [];
+            values = values.filter(v => !!v);
 
-            values.forEach((d, index) => {
+            for(let i = 0; i< order.length; i++){
+                sortedValues.push(values.find(o=> o.name === order[i]));
+            }
+
```
Golodhros: I would do something like:
```
let sortedValues = order.reduce((acc, current) => { 
    return [...acc, values.find(({name}) => name === current))]
},[]);
```
Or similar, I haven't tested it.

flacoman91: any idea why the js sort does that with more than 10 areas?

I made a separate codepen to only demonstrate the odd behavior of the sort.
https://codepen.io/flacoman91/pen/gBZKrz?editors=0010

src/charts/stacked-area.js
```diff
@@ -1019,11 +1019,11 @@ define(function(require){
             cleanDataPointHighlights();
 
             // ensure order stays constant
-            values = values
```
Golodhros: You still need the values here, right?

src/charts/stacked-area.js
Golodhros: Interesting! No idea though...

src/charts/stacked-area.js
flacoman91: doesn't seem like it. It works without it.

Golodhros: OK then

Golodhros:
Thanks for this fix @flacoman91 !!

Issue #692, opened 2018-12-14
Add hasLastHoverSliceHighlighted to donut chart


bootleq:
Hi, the donut's `hasFixedHighlightedSlice()` seems always highlight the slice set by `highlightSliceById()` initially, just good like the document says.

However, if we change the `highlightedSliceId` (for example, by using `highlightSliceById(4)` in a `customMouseOver` event), I would expect the fixed slice also change to it, while current behavior is keeping highlight the original slice.

This would be useful when we want to keep highlighting the last selected slice (after mouse out).

Version used: 2.9.8

Golodhros:
Hi @bootleq, thanks for opening an issue!

Oh, I am curious, how are you doing to update the highlighted slice within a customMouseOver? Could you post a Codepen reproducing this behavior?

You could use this one as a jumpstart: https://codepen.io/Britecharts/pen/WZRMEp

bootleq:
Thank you very much kindly provide a jumpstart!

What I tried is to call `highlightSliceById()` in customMouseOver, like codepen [bootleq/pen/wRGrPR](https://codepen.io/bootleq/pen/wRGrPR).
Trying to keep some slice highlighted, with respect the user's latest hover one.

Golodhros:
Hi @bootleq, 
I see the confusion. In Britecharts, after changing the settings of a chart, we need to call it again on the same container. In your case, it means doing this:

```
  donutChart
    .isAnimated(false)
    .highlightSliceById(1)
    .width(containerWidth)
    .height(containerWidth)
    .externalRadius(containerWidth/2.5)
    .internalRadius(containerWidth/5)
    .hasFixedHighlightedSlice(true)
    .on('customMouseOut', function(slice) { 
      console.log("Trying set fixed slice", slice.data.id);
      donutChart.highlightSliceById(slice.data.id);
      donutContainer.call(donutChart);
    });
```

Note how I had to remove the initial animation, as we are re-rendering the whole chart after each mouseOut. This isn't ideal.

I think what you are looking for is a new option that could be named 'hasLastHoverSliceHighlighted' or similar.

bootleq:
Thanks for explanation, now I understand the design.

I'm okay to close this issue, and besides, I found my original intention is to "keep last slice highlighted", similar to #365 with little difference, would be great if new option can be added.

DalerAsrorov:
This can be closed.

PR #694, opened 2018-12-17
Feature - last highlighted slice on mouse out transition

DalerAsrorov:
## Description

Added some additional logic for the new property `hasLastHoverSliceHighlighted`, which highlights the slice after `mouseout` event is called. *Note:* if donut char has both `hasFixedHighlightedSlice` and `hasLastHoverSliceHighlighted` as `true`, then `hasLastHoverSliceHighlighted` becomes the dominant factor. Otherwise, what's the point of setting it `true` (right?).

## Motivation and Context

https://github.com/eventbrite/britecharts/issues/692

## How Has This Been Tested?

Manual tests:
last_highlighted_slice

## Types of changes

## Checklist:

### DIFF
--- a/demos/src/demo-donut.js
+++ b/demos/src/demo-donut.js
@@ -36,7 +36,6 @@ function createDonutChart(optionalColorSchema) {
 
         donutChart
             .isAnimated(true)
-            .hasLastHoverSliceHighlighted(true)
             .highlightSliceById(2)
             .width(containerWidth)
             .height(containerWidth)
--- a/demos/src/demo-donut.js
+++ b/demos/src/demo-donut.js
@@ -36,6 +36,7 @@ function createDonutChart(optionalColorSchema) {
 
         donutChart
             .isAnimated(true)
+            .hasLastHoverSliceHighlighted(true)
             .highlightSliceById(2)
             .width(containerWidth)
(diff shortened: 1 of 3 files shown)


DalerAsrorov:
No problem! I just added setter/getter tests, but the DOM tests with Karma/Jasmine do not work if I try to check for the difference between `d` attributes before and after `mouseover` event, something like:
```javascript
            describe('when hasLastHoverSliceHighlighted is true', () => {

                it('hovered dounut chart slice should remain highlighted on mouseout', () => {
                    donutChart.hasLastHoverSliceHighlighted(true);
                    containerFixture.datum(dataset).call(donutChart);

                    let firstSlice = containerFixture.select('.chart-group .arc path');
                    let firstSliceNode = firstSlice.node();
                    let beforeAttr = firstSliceNode.getAttribute('d');
                    let afterAttr = null;

                    donutChart.on('customMouseOver', () => {
                        afterAttr = firstSliceNode.getAttribute('d');
                        // fails because the d value does not change
                        expect(beforeAttr).not.toBe(afterAttr);
                    });
                    firstSlice.dispatch('mouseover');
                });
            });
```
There is no example of a test that inspects slices' states before and after `mouseover` anywhere though.

PR #708, opened 2019-04-12
Fix: Fixing sparkline loading state, show it on demo and adding tests

Golodhros:
## Description
Fixes the loading state of the sparkline chart.
It also adds it to the demo and adds tests to line, stacked area and sparkline loading states.

## Motivation and Context

Demo wasn't showing the loading state, and it was broken.

## How Has This Been Tested?

Added tests

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/charts/sparkline.js
+++ b/src/charts/sparkline.js
@@ -8,9 +8,9 @@ define(function(require){
     const d3Selection = require('d3-selection');
     const d3Transition = require('d3-transition');
 
-    const {exportChart} = require('./helpers/export');
+    const { exportChart } = require('./helpers/export');
     const colorHelper = require('./helpers/color');
-    const {line} = require('./helpers/load');
+    const { stackedArea } = require('./helpers/load');
     const {uniqueId} = require('./helpers/number');
 
     const DEFAULT_TITLE_TEXT_STYLE = {
@@ -69,7 +69,7 @@ define(function(require){
             },
             width = 100,
             height = 30,
-            loadingState = line,
+            loadingState = stackedArea,
 
             xScale,
             yScale,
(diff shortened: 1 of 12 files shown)


src/charts/sparkline.js
```diff
@@ -8,9 +8,9 @@ define(function(require){
     const d3Selection = require('d3-selection');
     const d3Transition = require('d3-transition');
 
-    const {exportChart} = require('./helpers/export');
+    const { exportChart } = require('./helpers/export');
     const colorHelper = require('./helpers/color');
-    const {line} = require('./helpers/load');
+    const { stackedArea } = require('./helpers/load');
```
DalerAsrorov: Seems a bit confusing to leave the name of the loader as `stackedArea`. I would recommend something like:
```javascript
const { stackedArea: stackedAreaLoader } = require('./helpers/load');
```
Just to make it clear that it's a loading state not the chart itself.

src/charts/stacked-bar.js
```diff
@@ -14,9 +14,9 @@ define(function(require){
     const assign = require('lodash.assign');
     const d3Transition = require('d3-transition');
 
-    const {exportChart} = require('./helpers/export');
+    const { exportChart } = require('./helpers/export');
     const colorHelper = require('./helpers/color');
-    const {bar} = require('./helpers/load');
+    const { bar } = require('./helpers/load');
```
DalerAsrorov: Maybe an overkill but probably same here too

src/charts/sparkline.js
Golodhros: I was super close to do that, you just gave me the final push!
Thanks!

src/charts/stacked-bar.js
Golodhros: +1

PR #711, opened 2019-04-20
Allowing no quantities and adding demo on donut page

Golodhros:
## Description
Adds the possibility of rendering a vertical legend without quantities
Also added demo in the donut chart demo page

## Motivation and Context
Sometimes we just want to show a legend in a vertical format with names and colors. This is especially useful for time series charts

## How Has This Been Tested?
Added new data and tests

Tutorial__Donut_Chart

## Types of changes

### DIFF
--- a/demos/src/demo-donut.js
+++ b/demos/src/demo-donut.js
@@ -107,6 +107,28 @@ function getInlineLegendChart(dataset, optionalColorSchema) {
     }
 }
 
+function getVerticalLegendChart(dataset, optionalColorSchema) {
+    let legendChart = legend(),
+        legendContainer = d3Selection.select('.js-vertical-legend-no-quantity-container'),
+        containerWidth = legendContainer.node() ? legendContainer.node().getBoundingClientRect().width : false;
+
+    if (containerWidth) {
+        d3Selection.select('.js-vertical-legend-no-quantity-container .britechart-legend').remove();
+
+        legendChart
+            .width(containerWidth)
+            .height(100);
+
+        if (optionalColorSchema) {
+            legendChart.colorSchema(optionalColorSchema);
+        }
+
+        legendContainer.datum(dataset).call(legendChart);
+
+        return legendChart;
+    }
+}
+
 function createSmallDonutChart() {
     let donutChart = donut(),
         donutContainer = d3Selection.select('.js-small-donut-chart-container'),
@@ -131,7 +153,14 @@ function createSmallDonutChart() {
 }
 
 function createDonutWithHighlightSliceChart() {
-    let donutChart = donut(),
+    let dataNoQuantity = JSON.parse(JSON.stringify(datasetWithThreeItems))
+        .map((item) => {
+            delete item.quantity;
+
+            return item;
+        }),
+        legendChart = getVerticalLegendChart(dataNoQuantity),
+        donutChart = donut(),
         donutContainer = d3Selection.select('.js-donut-highlight-slice-chart-container'),
         containerWidth = donutContainer.node() ? donutContainer.node().getBoundingClientRect().width : false;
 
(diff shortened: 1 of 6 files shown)


demos/src/demo-donut.js
```diff
@@ -142,8 +171,15 @@ function createDonutWithHighlightSliceChart() {
             .width(containerWidth)
             .height(containerWidth/1.8)
             .externalRadius(containerWidth/5)
-            .internalRadius(containerWidth/10);
+            .internalRadius(containerWidth/10)
+            .on('customMouseOver', function (data) {
```
DalerAsrorov: It seems a bit weird with `data.data.id`, maybe renaming the callback argument as `response` or `payload` would help.

src/charts/legend.js
```diff
@@ -202,16 +204,53 @@ define(function(require){
...
+            }
+        }
+
+        /**
+         * Makes sure the types of the data are right and checks if it has quantities
+         * @param {LegendChartData} data
+         * @private
+         */
+        function cleanData(data) {
+            hasQuantities = data.filter(hasQuantity).length === data.length;
```
DalerAsrorov: Should there some kind of exception or error be thrown if some of the data has no quantity? In this case the equality wouldn't hold true.

src/charts/legend.js
```diff
@@ -202,16 +204,53 @@ define(function(require){
...
+
+            return data
+                .reduce((acc, d) => {
+                    if (d.quantity !== undefined) {
```
DalerAsrorov: What if `d.quantity` is a `null`? Could be checking for both `if(d.quantity !== undefined && d.quantity !== null)...`

src/charts/legend.js
```diff
@@ -202,16 +204,53 @@ define(function(require){
...
+                        d.quantity = +d.quantity;
```
DalerAsrorov: Is this operation done to make an item a `NaN`?

src/charts/legend.js
```diff
@@ -202,16 +204,53 @@ define(function(require){
...
+                    }
+                    d.name = String(d.name);
+                    d.id = +d.id;
```
DalerAsrorov: for this one as well: not sure what the `+` sign in front of the variable is supposed to do

src/charts/legend.js
```diff
@@ -333,17 +372,21 @@ define(function(require){
...
-              .append('text')
-                .classed('legend-entry-value', true)
-                .text(getFormattedQuantity)
-                .attr('x', chartWidth - valueReservedSpace)
-                .style('font-size', `${textSize}px`)
-                .style('letter-spacing', `${numberLetterSpacing}px`)
-                .style('text-anchor', 'end')
-                .style('startOffset', '100%');
+            if (hasQuantities) {
+                svg.select('.legend-group')
```
DalerAsrorov: I think it would make sense to transform the section of code in the `hasQuantities` statement to a function with a name that describes what this section of code is doing. That way it makes it consistent since the other `else` case does that.

demos/src/demo-donut.js
Golodhros: Good call, changing it

src/charts/legend.js
Golodhros: Well, the idea is to allow to pass data without quantity, so it renders what you see in the screenshot.
It makes sense on the time series charts

src/charts/legend.js
Golodhros: Makes sense!

src/charts/legend.js
Golodhros: No, to cast it into a number

src/charts/legend.js
Golodhros: Cleaner, yes. Done

Issue #258, opened 2017-05-23
Line chart: trace missing values with dashed path


evb-sun:
Related to https://github.com/eventbrite/britecharts/issues/226

Line charts are generally used to illustrate a trend over time, the chart should be able to tolerate missing values in time series.

## Expected Behavior
By default, the line chart should interpolate missing values, the line that represents missing values should visually differentiate from the rest of the chart:

<img width="722" alt="screen shot 2017-05-22 at 16 58 59" src="https://cloud.githubusercontent.com/assets/13459095/26333100/fde09d9c-3f0f-11e7-9b24-075e5acb3ee5.png">

Alternatively, instead of defaulting missing value to "0", we should simply display incomplete data:

<img width="789" alt="screen shot 2017-05-22 at 17 00 18" src="https://cloud.githubusercontent.com/assets/13459095/26333119/2cac8bf4-3f10-11e7-8a18-6f9e851efc0e.png">

**Please refer to Invision for design specs**
https://eventbrite.invisionapp.com/d/main/#/console/9951901/212621488/preview

## Current Behavior
Missing value is set to 0, resulting a "nose dive" look on the chart.

crozet-magenta:
also looking for this feature

Golodhros:
#741 Adds the support for empty spaces on the line chart.

Dashed lines would be ideal, but for the time being, we won't go that way.

Feel free to reopen this issue if that is necessary.

PR #741, opened 2019-07-08
Add support for null data point in line chart #258

shayh:
nulls are now handles correctly extending the time axis but not being drawn as point in the graph.

## How Has This Been Tested?
tested this change in my production environment handling null values

image

## Types of changes

- [] Refactor (changes the way we code something without changing its functionality)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] No linting issues
- [ ] Build is successful

### DIFF
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -34,6 +34,7 @@ define(function(require){
         uniqueId
     } = require('./helpers/number');
 
+    const acceptNullValue = (value) => value === null ? null : +value;
     /**
      * @typedef D3Selection
      * @type {Array[]}
@@ -596,10 +597,12 @@ define(function(require){
             const normalizedDataByTopic = dataByTopic.reduce((accum, topic) => {
                 let {dates, ...restProps} = topic;
 
-                let newDates = dates.map(d => ({
-                       date: new Date(d[dateLabel]),
-                       value: +d[valueLabel]
-                }));
+                let newDates = dates.map(d => {
+                    return {
+                        date: new Date(d[dateLabel]),
+                        value: acceptNullValue(d[valueLabel])
+                    };
+                });
 
                 accum.push({ dates: newDates, ...restProps });
 
@@ -711,6 +714,7 @@ define(function(require){
             topicLine = d3Shape.line()
                 .curve(curveMap[lineCurve])
                 .x(({date}) => xScale(date))
+                .defined(({value}) => value !== null)
                 .y(({value}) => yScale(value));
 
             lines = svg.select('.chart-group').selectAll('.line')


src/charts/line.js
```diff
@@ -598,7 +598,7 @@ define(function(require){
 
                 let newDates = dates.map(d => ({
                        date: new Date(d[dateLabel]),
-                       value: +d[valueLabel]
+                       value: d[valueLabel] === null ? null : +d[valueLabel]
```
DalerAsrorov: Do you mind creating a variable above `newDates` and assign this little logic to that variable?
```
const value = d[valueLabel] === null ? null : +d[valueLabel];
...
({
date: new Date(d[dateLabel]).
value
})
```

src/charts/line.js
```diff
@@ -711,6 +711,7 @@ define(function(require){
             topicLine = d3Shape.line()
                 .curve(curveMap[lineCurve])
                 .x(({date}) => xScale(date))
+                .defined(({value}) => value !== null)
```
DalerAsrorov:
```
.defined(({value}) => value)
```
should also do the trick

Golodhros: Wouldn't be instead
``` .defined(({value}) => !!value)```

File: src/charts/line.js (hunk 1 above)
Comment (shayh) [reply src/charts/line.js]: d is only defined within the map.
I don't see the advantage but I can do:
```
                let newDates = dates.map(d => {
                    const value = d[valueLabel] === null ? null : +d[valueLabel];
                    return {
                        date: new Date(d[dateLabel]),
                        value
                    };
                });
```

Comment (Golodhros) [reply src/charts/line.js]: What if we have an 'acceptNullValue' function?

const acceptNullValue = (value) => value === null ? null : value;

File: src/charts/line.js (hunk 2 above)
Comment (Golodhros) [reply src/charts/line.js]: (sorry, misclick)

Comment (shayh) [reply src/charts/line.js]: not if value is 0 which is a valid number but will produce false

File: src/charts/line.js (hunk 1 above)
Comment (shayh) [reply src/charts/line.js]: we still need to + the value to produce a valid number

Comment (Golodhros) [reply src/charts/line.js]: Oh, yeah, let's add that too

File: src/charts/line.js (hunk 2 above)
Comment (Golodhros) [reply src/charts/line.js]: True! Let's leave it like this then

File: src/charts/line.js (hunk 1 above)
Comment (shayh) [reply src/charts/line.js]: done

File: src/charts/line.js [hunk 3]
```diff
@@ -596,10 +596,13 @@ define(function(require){
             const normalizedDataByTopic = dataByTopic.reduce((accum, topic) => {
                 let {dates, ...restProps} = topic;
 
-                let newDates = dates.map(d => ({
-                       date: new Date(d[dateLabel]),
-                       value: +d[valueLabel]
-                }));
+                const acceptNullValue = (value) => value === null ? null : +value;
```
Golodhros: Oh, sorry, could you move this out of the function and to the top of the file?
That way we won't be recreating the function on each step of the reduce loop.

shayh: sure

shayh:
@Golodhros , @DalerAsrorov 
please let me know if anything else is needed to complete this PR

Golodhros:
Nothing! Thanks a lot @shayh for your contribution!

Issue #743, opened 2019-07-11
Move Britecharts into ES modules


Golodhros:
## Expected Behavior
We should have ES modules in our code so we can unblock features like tree shaking and because AMD is a format not used anymore by the community

## Possible Implementation
We need to transform our source code into ES modules

We have a half-done branch in https://github.com/Golodhros/britecharts/tree/ref-es6, having some issues with Karma that we can fix using the code on  https://github.com/Golodhros/pro-d3-building

PR #750, opened 2019-08-04
Refactor project into ES2015 modules

Golodhros:
## Description

Moves us out of AMD modules and into ES Modules for our charts, tests and demos

## Motivation and Context

#743 

## How Has This Been Tested?

Updated tests and manual on demos

## Types of changes

## Checklist:

- [ ] No linting issues
- [ ] Build is successful


demos/src/demo-grouped-bar.js
```diff
@@ -1,22 +1,20 @@
-'use strict';
+import { select, selectAll } from 'd3-selection';
+import { default as PubSub } from 'pubsub-js';
```
gagoar: this is odd,  es6 by default will export on 'default ' when you do export default. 

meaning, `import PubSub from 'pubsub-js'` should have worked.

demos/src/demo-kitchen-sink.js
```diff
@@ -1,26 +1,29 @@
...
-    loadScript('/britecharts/scripts/demo-stacked-bar');
-    loadScript('/britecharts/scripts/demo-grouped-bar');
-    loadScript('/britecharts/scripts/demo-donut');
-    loadScript('/britecharts/scripts/demo-line');
-    loadScript('/britecharts/scripts/demo-sparkline');
-    loadScript('/britecharts/scripts/demo-step');
-    loadScript('/britecharts/scripts/demo-brush');
-    loadScript('/britecharts/scripts/demo-scatter-plot');
-    loadScript('/britecharts/scripts/demo-bullet');
+const demoPaths = [
```
gagoar: if we are using es6 this might be better suited as a json file. 
and as a perk you get to add routes without changing code. (only the json file)

demos/src/demo-line.js
```diff
@@ -247,16 +244,16 @@ function isInRange(d0, d1, d) {
...
     createLineChart();
     createBrushChart();
     createLineChartWithSingleLine();
     createLineChartWithFixedHeight();
     createLoadingState();
 
     redrawCharts = function(){
-        d3Selection.selectAll('.line-chart').remove();
-        d3Selection.selectAll('.brush-chart').remove();
+        selectAll('.line-chart').remove();
```
gagoar: I was wondering , can't you do `selectAll('.line-chart, .brush-chart').remove()` ?

demos/src/demo-step.js
```diff
@@ -1,28 +1,26 @@
...
+import step from './../../src/charts/step';
+import miniTooltip from './../../src/charts/mini-tooltip';
+import { StepDataBuilder } from './../../test/fixtures/stepChartDataBuilder';
 
-const step = require('./../../src/charts/step');
-const miniTooltip = require('./../../src/charts/mini-tooltip');
-
-const dataBuilder = require('./../../test/fixtures/stepChartDataBuilder');
-let redrawCharts;
+require('./helpers/resizeHelper');
```
gagoar: interesting, is this a mokey patch ?

demos/src/demo-grouped-bar.js
Golodhros: True, changing. Thanks!

demos/src/demo-kitchen-sink.js
Golodhros: Makes sense, changing it. Thanks!

demos/src/demo-line.js
Golodhros: I can! Good catch!

demos/src/demo-step.js
Golodhros: It only triggers the 'resize' event on the PubSub module, so the rest of demos react to it.

Golodhros:
I am not sure. We aren't really changing anything from what we provide in our dist folder, so in theory, it is only a patch!

PR #777, opened 2019-12-08
Adds locale as a chainable operator to bar chart

maikdiepenbroek:
to enable passing a custom locale. This can be done in two forms: string (following the pattern: [a-z]{2}-[A-Z]{2} or an object in the form of a d3-format locale definition, f.e.: https://cdn.jsdelivr.net/npm/d3-format@1.4.2/locale/en-US.json. If a string is passed, it's locale is retrieved asynchronous and then passed to d3-format.formatDefaultLocale


src/charts/bar.js
```diff
@@ -164,7 +166,16 @@ define(function(require) {
          * @param {BarChartData} _data The data to attach and generate the chart
          */
         function exports(_selection) {
-            _selection.each(function(_data) {
+            if (locale && enableLabels) {
+                setDefaultLocale(locale)
+                    .then(() => initChart(_selection))
```
DalerAsrorov: Can initialization of the locale be done somewhere else?

src/charts/helpers/locale.js
```diff
@@ -0,0 +1,29 @@
+define(function (require) {
+    const {FORMAT_LOCALE_URL} = require('./constants');
+    const {formatDefaultLocale} = require('d3-format');
```
DalerAsrorov: There needs to be a test for this helper. Make sure that every use-case is covered.

src/charts/bar.js
maikdiepenbroek: Sure, i first had this in the drawLabels call. What place suits you better ?

src/charts/helpers/locale.js
maikdiepenbroek: Just to provide some context, this PR in its current state is only to discuss the current approach. If we decide it is the way to go, i can add it to the other charts and update docs/tests

maikdiepenbroek:
@DalerAsrorov and @Golodhros As a side note, when i perform the linting script i get a lot of code-style changes, also i see a lot of unused variables coming from require etc. When this PR is in a later state, i'll create a separate commit that fixes linting issues etc.

src/charts/bar.js
```diff
@@ -164,7 +166,16 @@ define(function(require) {
...
+            } else {
+                initChart(_selection);
+            }
+        }
+
+        function initChart(selection) {
+            selection.each(function(_data) {
```
Golodhros: I would say to trigger the locale request here and keep a reference to the promise. Something like:
```
  localeRequest = fetchLocale();
  selection.each(....
```
Later on, whenever the locale is actually needed, we'll do:
```
  localeRequest.then(...
```

That way we'll trigger the request as soon as possible (if needed) while blocking the render of anything that depends on it. 

We should probably include a configurable timeout in case the request fails.

maikdiepenbroek: Liked your suggestion, i've implemented it like this.

maikdiepenbroek:
@Golodhros and @DalerAsrorov  If you've got the time please review the current state of this PR. So far i've added jsdoc and unit tests for the locale helper. Also i added a small unit test for the locale property in the bart chart. Next to that i've implemented the suggestion of storing a reference to the getLocale promise. 
I'm having a difficult time getting to know the way of testing the complete bar chart, maybe you could help me with rounding of the tests especially mocking/spying. 

Once the bar chart is tested to standards,  i can move on and implement the locale property in the other charts (where applicable) and expand the docs.

Sounds like a plan ?

src/charts/bar.js
```diff
@@ -9,13 +9,13 @@ define(function(require) {
     const d3Format = require('d3-format');
     const d3Scale = require('d3-scale');
     const d3Selection = require('d3-selection');
-    const d3Transition = require('d3-transition');
```
Golodhros: This needs to be loaded, even if we don't use it.

src/charts/bar.js
```diff
@@ -1292,6 +1302,15 @@ define(function(require) {
             return this;
         };
```
Golodhros: We would need a comment on this accessor so we generate the documentation

src/charts/bar.js
```diff
@@ -177,7 +183,11 @@ define(function(require) {
                 drawBars();
                 drawAxis();
 
-                if (enableLabels) {
+                if (enableLabels && locale) {
+                    localeRequest
```
Golodhros: Would it work if we set up localeRequest as a resolved promise by default?

src/charts/helpers/locale.js
```diff
@@ -0,0 +1,44 @@
...
+    const { formatDefaultLocale } = require('d3-format');
+    const { json } = require('d3');
+    /**
+     * Sets the given locale as the new default locale through d3-format's formatDefaultLocale
+     * When a string is given as a parameter, it uses d3.json to fetch the correct locale definition.
+     * When an object is used, it simply uses it to set the new locale.
+     * @param  {string | Object} locale The locale in string or object format.
+     * @return {Promise}       Result of formatDefaultLocale call.
+     */
+
```
Golodhros: Nit: No space here, thanks!

src/charts/helpers/locale.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Checks if a locale is of the correct format
+     * @param  {string}  locale Locale to check
+     * @return {Boolean}       Is the locale has a correct format (https://cdn.jsdelivr.net/npm/d3-format/locale/)
+     */
+
+    function isValidLocale(locale) {
+        return /[a-z]{2}-[A-Z]{2}/.test(locale);
```
Golodhros: I would move the RegEx as a constant at the top of this file.

src/charts/helpers/locale.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    function isValidLocaleDefinition(locale) {
+        const requiredKeys = ['decimal', 'thousands', 'grouping', 'currency']
```
Golodhros: I would move it to a constant at the top of the file as well.

src/charts/helpers/locale.js
```diff
@@ -0,0 +1,44 @@
...
+     * Checks if a locale is of the correct format
+     * @param  {string}  locale Locale to check
+     * @return {Boolean}       Is the locale has a correct format (https://cdn.jsdelivr.net/npm/d3-format/locale/)
+     */
+
+    function isValidLocale(locale) {
+        return /[a-z]{2}-[A-Z]{2}/.test(locale);
+    }
+
+    function isValidLocaleDefinition(locale) {
```
Golodhros: Missing a small comment here, thanks!

test/specs/bar.spec.js
```diff
@@ -767,6 +767,18 @@ define(['d3', 'bar', 'barChartDataBuilder'], function(d3, chart, dataBuilder) {
...
+
+                barChart.locale(testLocale);
+                newLocale = barChart.locale();
+
+                expect(defaultLocale).not.toBe(newLocale);
+                expect(newLocale).toBe(testLocale);
+            })
         });
     });
 });
```
Golodhros: I would add another test within the 'render' section, checking that a label is different when using a different locale.

test/specs/helpers.spec.js
```diff
@@ -441,5 +443,34 @@ define([
...
+                expectAsync(locale.setDefaultLocale(validLocaleDefinition)).toBeResolved();
+                done();
+            });
+
+            it('should return a rejected promise when a invalid locale definition is given', (done) => {
+                const invalidLocaleDefinition = {};
+                expectAsync(locale.setDefaultLocale(invalidLocaleDefinition)).toBeRejected();
+                done();
+            });
+        });
```
Golodhros: I would add a couple of test cases for the two OK cases.

package.json
```diff
@@ -80,6 +80,7 @@
     "babel-eslint": "7.2.3",
     "babel-istanbul": "^0.12.2",
     "babel-loader": "7.1.2",
+    "babel-polyfill": "6.26.0",
```
Golodhros: Why did you need this? Only for the tests?

src/charts/bar.js
maikdiepenbroek: Got it, i reversed it.

src/charts/bar.js
maikdiepenbroek: Added it

src/charts/helpers/locale.js
maikdiepenbroek: Removed it.

src/charts/helpers/locale.js
maikdiepenbroek: Moved it to a const.

src/charts/helpers/locale.js
maikdiepenbroek: Moved it to a const

src/charts/helpers/locale.js
maikdiepenbroek: Added it

test/specs/helpers.spec.js
maikdiepenbroek: I had those tests, but since i wasn't able to mock the d3.json call, i removed those tests since i would be relying on an internet connection to make the test pass. If that's not a problem, i could re add the OK cases test.

package.json
maikdiepenbroek: This was required for the tests indeed, the parsing failed due to missing Promise support.

src/charts/bar.js
maikdiepenbroek: I'm not sure what you mean, could you explain with some (pseudo) code ?

test/specs/bar.spec.js
maikdiepenbroek: I'm having trouble getting the test to pass, a little help here would be very much appreciated.

src/charts/bar.js
Golodhros: Did you push the changes?

src/charts/bar.js
Golodhros: Instead of declaring localeRequest as undefined, you could do:
  localeRequest = Promise.resolve();

Then, you wouldn't need to add the branching here:

   localeRequest.then(() => drawLabels())
         .catch(() => new Error('You need an Internet connection in order to load the selected locale'))

test/specs/bar.spec.js
Golodhros: OK, could you upload the test in a commented state?

test/specs/helpers.spec.js
Golodhros: Makes sense, add those here commented for now, I will take a look later.

src/charts/bar.js
maikdiepenbroek: No, did the changes locally only.

test/specs/bar.spec.js
maikdiepenbroek: Sure, coming up!

test/specs/helpers.spec.js
maikdiepenbroek: Ok, re-added them

src/charts/bar.js
maikdiepenbroek: I've added the Promise.resolve to the declaration of localeRequest, but i don't quite understand the  branching part.

Golodhros: If the value of localeRequest is a resolved promise by default, then you don't need to check for 'enableLables' or 'locale', you can just put the same code for all cases.

maikdiepenbroek: > If the value of localeRequest is a resolved promise by default, then you don't need to check for 'enableLables' or 'locale', you can just put the same code for all cases.

Got it now, good catch!

test/specs/bar.spec.js
maikdiepenbroek: I've added the commented test, been driving me crazy haha. I hope you can take a look. @Golodhros

src/charts/bar.js
```diff
@@ -1292,6 +1300,22 @@ define(function(require) {
             return this;
         };
 
+        /**
+         * Gets or Sets the locale which d3-format uses.
+         * (Default is en-US)
+         * @param  {string} _x                   Desired locale
```
Golodhros: You can do [_x = 'es-US'] to mark the default.

Could you add a link in the comment to point to the possible values?
Also, if we accept a locale object instead of the string, we should explain its values or add another link for referencing the format of it.

src/charts/bar.js
```diff
@@ -178,7 +184,9 @@ define(function(require) {
                 drawAxis();
 
                 if (enableLabels) {
-                    drawLabels();
+                    localeRequest
+                        .then(() => drawLabels())
```
Golodhros: Wouldn't the locale change also the axis tick text?

maikdiepenbroek: I'm not sure to be honest.

test/specs/bar.spec.js
```diff
@@ -176,6 +176,44 @@ define(['d3', 'bar', 'barChartDataBuilder'], function(d3, chart, dataBuilder) {
                     expect(actualOffset).toBe(expectedOffset);
                 });
             });
+
+            // describe('Locale', () => {
```
Golodhros: Nit: you can just add an 'x' as in 'xdescribe' to comment the whole thing.

maikdiepenbroek: True, but we agreed on pushing a commented out test for you to have a look at, right ?

Golodhros: Yeah, I mean that adding the 'x' is how you comment the test. Not a big deal though.

src/charts/bar.js
Golodhros: No prob, I will test it later.

src/charts/bar.js
maikdiepenbroek: Fixed

test/specs/bar.spec.js
maikdiepenbroek: Fixed

src/charts/bar.js
```diff
@@ -210,28 +218,28 @@ define(function(require) {
         function buildContainerGroups() {
             let container = svg
                 .append('g')
-                  .classed('container-group', true)
-                  .attr('transform', `translate(${margin.left + yAxisPaddingBetweenChart}, ${margin.top})`);
+                .classed('container-group', true)
```
Golodhros: Oh, what happened here?
append is a special case that gets 2 spaces instead of four.
Was this eslint?

maikdiepenbroek: This happened after i ran `yarn lint:fix`

Golodhros: I see, is there an easy way of rollback that change only? It messed up all the indentation.

maikdiepenbroek: Rolled back before the linting 👍

Golodhros:
Awesome work @maikdiepenbroek!!!

Let me play a bit with it and see how we move forward, thanks a lot!

maikdiepenbroek:
@Golodhros No problem, been fun and thanks for the patience and collaboration 👍

Golodhros:
> @Golodhros No problem, been fun and thanks for the patience and collaboration 👍

On the contrary! Thanks for your patience and sorry for the back and forths!

PR #781, opened 2019-12-11
Tuning up locale feature on Bar Chart

Golodhros:
## Description

Adjust default locale feature to use the configured format function

## Motivation and Context

Fixes the set default locale not being applied

## How Has This Been Tested?

Ran tests
More test will be needed

## Types of changes

## Checklist:

- [ ] No linting issues
- [ ] Build is successful


Golodhros:
Hey @maikdiepenbroek, here is a PR adjusting your work to apply the right formatting function (the one returned by d3's formatDefaultLocale function).

This is working properly as is, although there are some tests (yours and another couple) that stop working. This happens because all the code within the promise chain will run in a different call stack. I have tried several things with Jasmine but nothing worked.

Overall, I think this feature is really cool. However, the addition of the extra asynchronous call to fetch the locale is adding a good amount of complexity. I honestly think we could deal with it, but I am afraid it would give a lot of headaches and it will increase the maintenance work on the library if applied all over our charts.

Due to this increased complexity, I suggest to just include the option of setting the locale via an object. If a user needs to dynamically fetch the locale info, they can do it beforehand, and only then render the chart passing the locale object.

Would that work for your use case?

maikdiepenbroek:
Thanks for the touch up of the functionality! Looks good! I think that you're right. Seeing the amount of effort it takes to even test is properly is reason enough to take the async part out and only support the object notation. This way the applications that are implementing it can call the external service to fetch the correct locale themselves and just pass the object. Is that something you can implement or should i?

I already implemented the async way in my use case but I can change that to use the object notation.

Great suggestion and thanks for fixing up the things i missed

Golodhros:
Hi @maikdiepenbroek, I think this is ready to go and we can go ahead and use this same pattern on the rest of the charts.

I won't have time in the next weeks to do this for the other charts, but I would be able to do code reviews.

I am also thinking that maybe we could create a higher order function that wraps the chart and accepts the locale string, that way, we'll be able to seamlessly setup the chart and render it after getting back the locale object.

It could be something like:
```
        barChart
            .width(containerWidth)
            .hasPercentage(true)
            .enableLabels(true)
            .labelsNumberFormat('.0%')
            .height(300);

		renderWithLocale('es-ES', barContainer, dataset, barChart);
```

maikdiepenbroek:
Perfect, looks good. I've just changed my local build's implementation to use the locale object only so for my project / use case i'm good to go as is.

Since i would like to see it released a.s.a.p i'll implement this in the remaining charts and create a PR for you to review.

Cool suggestion regarding the higher order function but i like the current api / implementation better to be honest. What does your suggestion provide us, since i would say it's only an extra function for the user to call.

Golodhros:
Awesome @maikdiepenbroek, thanks a lot for your effort on this feature!

The higher order function would be a nice to have for people that have available just the locale string and would like a convenient way of not diving into the details of the locale. Not super important though, I will keep a branch around and see if I can put together something.

I will merge this branch, and once we merge all of the rest we can publish a new minor version of Britecharts.

Another option would be to start merging it into our alpha for the version 3. We have some exciting changes coming (ES modules among others).

Golodhros:
Hi @DalerAsrorov and @maikdiepenbroek,

I have spent a couple of hours fixing merge conflicts and getting things ready with our RC for version 3. You can see the branch [here](https://github.com/eventbrite/britecharts/tree/release/3.0.0-rc2).

It has been a bit painful, so I think that we must freeze our master branch and only include security updates or critical bugs. We will only develop against the RC branch.

I have updated the [v3 project on Github](https://github.com/eventbrite/britecharts/projects/2) to set the minimum for the release. I think it could be doable to do in a couple of months.

What are your thoughts?

maikdiepenbroek:
Oof, thanks for the effort! 

I agree that we should leave the master for bugs/security updates only.

For now i will develop the locale feature from the 3.0RC branch.

I do have one question, is it still possible to create a release once the locale feature is merged ? Then i can use that 3.0 RC version in my project.

Golodhros:
Awesome, thanks @maikdiepenbroek!

Yes, I think so. I will double-check, but I think we can release alpha versions that would be available on NPM. If not, you could always point to the branch using the GitHub URL, but that's a bit less handy.

maikdiepenbroek:
> Awesome, thanks @maikdiepenbroek!
> 
> Yes, I think so. I will double-check, but I think we can release alpha versions that would be available on NPM. If not, you could always point to the branch using the GitHub URL, but that's a bit less handy.

No problem! 
I'm not using the npm package in a package.json, but linking it directly from the cdn. So i hope that won't cause issues, but i'm sure it wont :)

Issue #338, opened 2017-08-24
Support negative values


peebeebee:
Almost all charts should be able to accept and display negative values, except maybe the Donutchart.

## Expected Behavior
When passing in a  negative value, it should go below the x/y-axis. 

## Current Behavior
Right now, we get an error in the console telling us:
Error: <rect> attribute width: A negative value is not valid. 

## Steps to Reproduce (for bugs)
Just pass in a negative value

JArmando:
Part of the fix could be in the `buildScales` function. There is a line setting the min Y axis value that will (in spirit) always assume that the min value is positive:
`let yScaleBottomValue = Math.abs(minY) < 0 ? Math.abs(minY) : 0;`
if you let it be `minY` then it "works". Only problem is that it doesn't handle the ticks correctly.

![image](https://user-images.githubusercontent.com/7327505/37616549-500efb4e-2b7e-11e8-8baa-24b9779f7576.png)

In this example I manually set `yScaleBottomValue  = -100`. I'll try and hack it for my particular case, might dedicate some time to propose a better solution.

Golodhros:
Hi @JArmando, thanks for your comment!

We would love to see your solution to this issue and merge it on Britecharts!

Golodhros:
@cherurg graciously added this feature for the line chart.

We still have a bunch to go though!

jkohlin:
Being new at this. Is there any chance I can have negative values displayed correctly in a line chart today?

Golodhros:
Hi @jkohlin, 
There was an error on the last release build, I will take care of it so that you can use this feature!

Golodhros:
Sorry about the delay.
The fixes are on the latest britecharts release, but we are having issues with the publishing on NPM. I am getting help from our OPS team to figure out what's going on on that publishing pipeline.

jkohlin:
I see. Well no need to hurry for me anymore. I had to move on and find a charting library that had support for this. Thanks anyway for responding.

miglesiasEB:
We managed to publish again on NPM, Britecharts 2.14.4 is live with this change for the line chart!

Golodhros:
@ajdani implemented this for the stacked area chart on #799

Golodhros:
The missing charts are:
- All bar charts (regular, horizontal, stacked and grouped)
- Brush Chart
- Scatter Plot

PR #799, opened 2020-04-11
Adding support for negative area charts / allow scaling of y-axis below 0

ajdani:
Added support for _negative_ area charts (meaning that the stacked area reaches below y-values of 0) - in general the solution is really similar to the solution which was implemented for the line-chart as well some weeks ago. Highlighting the 0-axis in case of negative values is implemented as well, exactly as it was done by @cherurg in the line-chart feature to be consistent.

To avoid confusion:
In this context I speak of a _negative_ area chart, if the stacked-area is below 0. 

## Motivation and Context
This feature is missing until now, there is need for it - but currently there is no issue for it.

## How Has This Been Tested?

- Check if the first tick is negative in case of a _negative_ area chart
- Check that there exists the `horizontal-grid-line--highlighted` class in case of a _negative_ area chart
- Check that this class is _not_ there, in case of a _non-negative_ area chart

image

## Types of changes

## Checklist:
- [x] No linting issues
- [x] Build is successful


Golodhros:
Related to https://github.com/eventbrite/britecharts/issues/338

PR #804, opened 2020-04-14
Add yAxisBaseline to the stacked-area chart

ajdani:
## Description
Adds the functionality to set the yAxisBaseline to a custom-value for the stacked-area chart. This baseline is the y-value where the stacked area is starting from.

Setting `.yAxisBaseline(-20)` sets the baseline to **-20** and changes the _minY_ of the scale as well to -20.

## How Has This Been Tested?
- Test that the maxY is set to baseline if the baseline is bigger then the max of the data
- Test that the minY is set to baseline if the baseline is smaller then the min of the data
- Test the getter / setter of the yAxisBaseline

tbh there is no test which ensures that the drawing is correct, e.g. the baseline changes. I did not have an idea how to tackle this. Maybe you can provide some ideas please?

image

## Types of changes

## Checklist:
- [ ] No linting issues
- [x] Build is successful


src/charts/stacked-area.js
```diff
@@ -1561,6 +1584,22 @@ define(function(require){
             return this;
         };
 
+        /**
+         * Gets or Sets the yAxisBaseline
```
DalerAsrorov, line 1588: This is the part that goes into API docs. Could you add a more detailed description of what it does here? In addition, providing an example here would be super helpful as well.

DalerAsrorov: You can document an example of code like this:
https://github.com/eventbrite/britecharts/blob/master/src/charts/stacked-area.js#L1519

ajdani: Thanks for the hint how to document examples. Will add it.

ajdani:
Thanks for the review @DalerAsrorov !

Added some documentation and an example.

PR #805, opened 2020-04-14
Add numerical and logarithmic x-axis for stacked-area charts

ajdani:
Adds the functionality to use numerical values as x-values to plot, instead of only dates - this can be achieved by setting `.xAxisValueType('number')` - the default behaviour will be the same (so we introduce no breaking API changes!). 

The user can use the same method for the tooltip to change the formatting of the x-values in the tooltip-title.

Additionally you can modify the area-chart with `.xAxisScale('logarithmic')` to show a logarithmic scaled axis. 

## How Has This Been Tested?
- Test when xAxisValueType is set to numeric that the x-Axis ticks are not dates anymore but numbers
- Test when xAxisValueType is set to numeric and xAxisScale to logarithmic, that the x-Axis ticks are in the format `10^x` and not dates anymore
- Test the getter / setter for `xAxisValueType` and `xAxisScale` for the stacked-area chart
- Test the getter / setter for `xAxisValueType` for the tooltip
- Test when xAxisValueType is set to numeric for the tooltip, no dates are shown in the title, but the numerical value

image

image

## Types of changes

I did not want to introduce any breaking change in the API, so I did not refactor the variable-namings, etc. too much. This means that it might be strange now if you have variables with `date` in it but the value of the variable does not need to be a date anymore.

For me at least it's now consistent to the other chart-types and the refactoring of namings, etc. may be done in a bigger MR when rolling out these features for all charts?

## Checklist:
- [ ] No linting issues
- [x] Build is successful

### DIFF
--- a/src/charts/helpers/axis.js
+++ b/src/charts/helpers/axis.js
@@ -76,6 +76,18 @@ define(function(require) {
     }
 
     /**
+     * Calculates the maximum number of ticks for the x axis
+     * with respect to number ranges
+     * @param  {Number} width               Chart width
+     * @param  {Number} dataPointNumber     Number of entries on the data
+     * @return {Number}                     Number of ticks to render
+     */
+    const getMaxNumOfHorizontalTicksForNumberRanges = (width, dataPointNumber) => {
+        let ticksForWidth = Math.ceil(width / (singleTickWidth + horizontalTickSpacing));
+        return Math.min(dataPointNumber, ticksForWidth);
+    }
+
+    /**
      * Returns tick object to be used when building the x axis
      * @param {dataByDate} dataByDate       Chart data ordered by Date
      * @param {Number} width                Chart width
@@ -112,8 +124,32 @@ define(function(require) {
         };
     };
 
+    /**
+     * Returns tick object to be used when building the x axis
+     * @param {dataSorted} dataSorted       Chart data ordered by Date
+     * @param {Number} width                Chart width
+     * @param {String} [settings=null]      Optional forced settings for axis
+     * @return {object} tick settings for minor axis
+     */
+    const getSortedNumberAxis = (dataSorted, width) => {
+        const firstEntry = dataSorted[0].date;
+        const lastEntry = dataSorted[dataSorted.length - 1].date;
+        const timeSpan = lastEntry - firstEntry;
+
+        const minorTickValue = getMaxNumOfHorizontalTicksForNumberRanges(
+            width,
+            timeSpan
+        );
+
+        return {
+            tick: minorTickValue,
+            format: null
+        };
+    };
+
     return {
-        getTimeSeriesAxis
+        getTimeSeriesAxis,
+        getSortedNumberAxis
     };
 
 });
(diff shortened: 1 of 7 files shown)


src/charts/helpers/axis.js
```diff
@@ -112,8 +124,32 @@ define(function(require) {
...
+        const timeSpan = lastEntry - firstEntry;
+
+        const minorTickValue = getMaxNumOfHorizontalTicksForNumberRanges(
+            width,
+            timeSpan
+        );
+
+        return {
+            tick: minorTickValue,
+            format: null
```
DalerAsrorov, line 146: Is `format:null` necessary? As I understand, by default it will be `undefined`. Would `null` be a better option? Just curious why.

src/charts/stacked-area.js
```diff
@@ -313,28 +315,45 @@ define(function(require){
...
-                };
+            if(xAxisValueType === 'number') {
+                minor = getSortedNumberAxis(dataSorted, width);
                 major = null;
+
+                if(xAxisScale === 'logarithmic') {
+                    xAxis = d3Axis.axisBottom(xScale)
+                        .ticks(minor.tick, "e")
+                        .tickFormat(function (d) {
+                            var log = Math.log(d) / Math.LN10;
```
DalerAsrorov, line 326: Could you use const here?

src/charts/helpers/axis.js
ajdani: Oh no, just wanted to stick to the interface and be consistent. So I remove it completely, as it does not make sense.

src/charts/stacked-area.js
ajdani: Oh, yes sure - did not see that. Will update it.

ajdani:
Thanks for the review @DalerAsrorov !

I've fixed the issues mentioned by you and added some examples for the docs as well.

ajdani:
@DalerAsrorov 
Added the same functionality for line-charts as well. 

So you might only merge this one:
https://github.com/eventbrite/britecharts/pull/806

src/charts/stacked-area.js
```diff
@@ -464,21 +483,49 @@ define(function(require){
          * @private
          */
         function buildScales() {
+            xScale = buildXAxisScale();
+            yScale = buildYAxisScale();
```
Golodhros, line 487: Awesome!

src/charts/stacked-area.js
```diff
@@ -899,21 +946,31 @@ define(function(require){
...
                 });
+        }
+
+        /**
+         * Casts the data given to a date or number
+         * respecting the value of xAxisValueType
+         * @param {string | number} value   Value data
+         * @return {Date | number} value    Casted value
+         */
+        function castValuesToType(value) {
```
Golodhros, line 968: Great job

src/charts/helpers/axis.js
```diff
@@ -112,8 +124,31 @@ define(function(require) {
...
+        };
+    };
+
     return {
-        getTimeSeriesAxis
+        getTimeSeriesAxis,
+        getSortedNumberAxis
     };
 
 });
```
Golodhros, line 154: Could you add a couple of basic tests for these helpers?

Thanks!

ajdani:
@Golodhros I went for the other PR, lets close this one.

PR #806, opened 2020-04-17
Add numerical and logarithmic x-axis for line charts 

ajdani:
Adds the functionality to use numerical values as x-values to plot, instead of only dates - this can be achieved by setting .xAxisValueType('number') - the default behaviour will be the same (so we introduce no breaking API changes!).

The user can use the same method for the tooltip to change the formatting of the x-values in the tooltip-title.

Additionally you can modify the line-chart with .xAxisScale('logarithmic') to show a logarithmic scaled axis.

**Note: this PR is based on my other open PR, so it should be merged after the other or instead of the other one.**

Additionally fixes a bug where in case of a line-chart with only negative values as yMax the `Math.abs(maxY)` was taken instead of only `maxY`.

## How Has This Been Tested?
- Test if yScale in case of a fully negative line chart is correct
- Test the getter / setter / default value of xAxisValueType, xAxisScale
- Test if x-Axis ticks are numbers in case of xAxisValueType === 'number'
- Test if x-Axis ticks are logarithmic numbers in case of xAxisScale === 'logarithmic'

image

image

## Types of changes

## Checklist:
- [ ] No linting issues
- [x] Build is successful
- [x] Updated the documentation


test/fixtures/stackedAreaDataBuilder.js
```diff
@@ -8,7 +8,8 @@ define(function(require) {
         jsonSalesChannel = require('../json/areaDataSalesChannel.json'),
         jsonReportService = require('../json/areaDataReportService.json'),
         jsonLargeService = require('../json/areaDataLarge.json'),
-        jsonNegativeValues = require('../json/areaDataNegativeValues.json');
+        jsonNegativeValues = require('../json/areaDataNegativeValues.json'),
+        jsonNumericKeys = require('../json/areaDataNumericKeys.json');
```
Golodhros, line 12: Thanks for these!

src/charts/helpers/type.js
```diff
@@ -0,0 +1,22 @@
...
+            return Number(value);
+        }
+
+        return new Date(value);
+    }
+
+    return {
+        castValueToType: castValueToType
+    };
+});
```
Golodhros, line 22: Looks good!

Could you create a small test for this helper?

ajdani:
Added tests and rebased on current master. @Golodhros

PR #807, opened 2020-04-19
Add support for multirow tooltip title

ajdani:
This PR adds support for text wrap of the tooltip title - currently the title + current date would just overlap the tooltip, but now the existing textWrap functionality is used for the title as well.

## Motivation and Context
Long tooltip titles + date formats break the style of the tooltip, so the solution is a text wrap for the title row as well.

## How Has This Been Tested?
- Test if the tooltipDivider changes the y-Position in case of a long title and check if the title is correct wrapped

Before:
image

After:
image

## Types of changes

## Checklist:
- [ ] No linting issues
- [x] Build is successful

### DIFF
--- a/src/charts/helpers/text.js
+++ b/src/charts/helpers/text.js
@@ -150,9 +150,41 @@ define(function(require) {
         return b.measureText(text).width;
     }
 
+    /**
+     * Heuristic which gets the number of lines needed to display the title of the tooltip
+     * If shouldShowDateInTitle is set to true, it takes the formatted Date.now() as additional influencer
+     * for the approximation of the needed number of lines.
+     * @param  {String}  text  Text which shall be tested for the necessary number of lines
+     * @param  {Number}  fontSize  Fontsize to use for the heuristic
+     * @param  {Number}  maxLength  Maximal length per line
+     * @return  {Number}  approximateLineNumber  Approximative number of lines needed to display the title
+     * @private
+     */
+    const getApproximateNumberOfLines = function(title, fontSize, maxLength) {
+        const words = title.split(/\s+/).reverse();
+
+        var word,
+            line = [],
+            approximateLineNumber = 1;
+
+        while (word = words.pop()) {
+            line.push(word);
+
+            const textWidth = getTextWidth(line.join(' '), fontSize, 'Karla, sans-serif');
+            if (textWidth > maxLength) {
+                line.pop();
+                line = [word];
+                ++approximateLineNumber;
+            }
+        }
+
+        return approximateLineNumber;
+    }
+
     return {
         getTextWidth,
         wrapText,
-        wrapTextWithEllipses
+        wrapTextWithEllipses,
+        getApproximateNumberOfLines
     };
 });
--- a/src/charts/helpers/text.js
+++ b/src/charts/helpers/text.js
@@ -163,11 +163,10 @@ define(function(require) {
     const getApproximateNumberOfLines = function(title, fontSize, maxLength) {
         const words = title.split(/\s+/).reverse();
 
-        var word,
-            line = [],
+        let line = [],
             approximateLineNumber = 1;
 
-        while (word = words.pop()) {
+        for(let word of words) {
             line.push(word);
 
             const textWidth = getTextWidth(line.join(' '), fontSize, 'Karla, sans-serif');
--- a/src/charts/tooltip.js
+++ b/src/charts/tooltip.js
@@ -418,7 +418,7 @@ define(function(require){
                 tooltipMaxTitleLength
             );
 
-            if(approximateNumberOfTitleLines > 1) {
+            if (approximateNumberOfTitleLines > 1) {
                 additionalTooltipTitleHeight = 17 * (approximateNumberOfTitleLines -1)
             }
         }
@@ -430,10 +430,10 @@ define(function(require){
          * @private
          */
         function updateTitle(dataPoint) {
-            const tTitle = getTooltipTitle(dataPoint[dateLabel]);
+            const textTitle = getTooltipTitle(dataPoint[dateLabel]);
 
             tooltipTitle
-                .text(tTitle)
+                .text(textTitle)
                 .call(textWrap, tooltipMaxTitleLength, getTooltipTitleXPosition());
         }
 
@@ -443,18 +443,18 @@ define(function(require){
          * @private
          */
         function getTooltipTitle(date) {
-            let tTitle = title;
+            let textTitle = title;
             let formattedDate = formatKey(date);
 
-            if (tTitle.length) {
+            if (textTitle.length) {
                 if (shouldShowDateInTitle) {
-                    tTitle = `${tTitle} - ${formattedDate}`;
+                    textTitle = `${textTitle} - ${formattedDate}`;
                 }
             } else {
-                tTitle = formattedDate;
+                textTitle = formattedDate;
             }
 
-            return tTitle;
+            return textTitle;
         }
 
         /**
--- a/src/charts/tooltip.js
+++ b/src/charts/tooltip.js
@@ -444,7 +444,7 @@ define(function(require){
          */
         function getTooltipTitle(date) {
             let tTitle = title;
-            let formattedDate = formatKey(dataPoint[dateLabel]);
+            let formattedDate = formatKey(date);
 
             if (tTitle.length) {
                 if (shouldShowDateInTitle) {
--- a/src/charts/tooltip.js
+++ b/src/charts/tooltip.js
@@ -14,7 +14,7 @@ define(function(require){
         isInteger
     } = require('./helpers/number');
 
-    const {getTextWidth} = require('./helpers/text');
+    const {getTextWidth, getApproximateNumberOfLines} = require('./helpers/text');
 
     /**
      * Tooltip Component reusable API class that renders a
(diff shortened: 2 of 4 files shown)


ajdani:
Oh my god, this is a flaky test - will fix it in the next days. Runs locally - seems to be dependent to the resolution of the Browser used in tests... :-(

Golodhros:
I see, sometimes it is useful to run 'yarn run test:travis' on dev to make sure is not something related to the environment!

src/charts/tooltip.js
```diff
@@ -400,15 +405,68 @@ define(function(require){
...
+        }
+
+        /**
+         * Heuristic which gets the number of lines needed to display the title of the tooltip
+         * If shouldShowDateInTitle is set to true, it takes the formatted Date.now() as additional influencer
+         * for the approximation of the needed number of lines.
+         * @return  {Number}  approximateLineNumber  Approximative number of lines needed to display the title
+         * @private
+         */
+        function getApproximateNumberOfLines() {
```
Golodhros, line 426: Maybe if you could make this function more generic we could add it to our text helper?

test/specs/tooltip.spec.js
```diff
@@ -122,6 +122,57 @@ define(['jquery', 'd3', 'tooltip'], function($, d3, tooltip) {
...
+                            .select('.tooltip-divider').attr('y1');
+
+                        actualDividerY2Position = containerFixture.select('.britechart-tooltip')
+                            .select('.tooltip-divider').attr('y2');
+
+                        expect(actualTitle).toEqual(expectedTitle);
+                        expect(parseInt(actualDividerY1Position)).toEqual(expectedDividerYPosition);
+                        expect(parseInt(actualDividerY2Position)).toEqual(expectedDividerYPosition);
+                    });
                 });
```
Golodhros, line 176: Thanks for the test!

src/charts/tooltip.js
ajdani: Moved it and added a test for it.

ajdani:
Rebased and moved the function to helpers and added a test for it.

> I see, sometimes it is useful to run 'yarn run test:travis' on dev to make sure is not something related to the environment!

Thanks for the hint. Somehow this runs without errors locally. Will see if the pipeline fails here.

ajdani:
Ok, pipeline fails - maybe we can have a look on this together if the results of `yarn run test:travis` should be deterministic we should fix the behavior of this rather than my failing test. @Golodhros @DalerAsrorov 

I'm running `node`, `npm` and `yarn` in newer version as the CI - do you think this may cause the differences? Maybe different resolved dependencies...?

````
node --version
v12.13.0
npm --version
6.12.0
yarn --version
1.19.1
```

src/charts/helpers/text.js
```diff
@@ -150,9 +150,41 @@ define(function(require) {
...
+     * @param  {String}  text  Text which shall be tested for the necessary number of lines
+     * @param  {Number}  fontSize  Fontsize to use for the heuristic
+     * @param  {Number}  maxLength  Maximal length per line
+     * @return  {Number}  approximateLineNumber  Approximative number of lines needed to display the title
+     * @private
+     */
+    const getApproximateNumberOfLines = function(title, fontSize, maxLength) {
+        const words = title.split(/\s+/).reverse();
+
+        var word,
```
DalerAsrorov, line 166: Sorry didn't see it before, please use `let`

src/charts/helpers/text.js
```diff
@@ -150,9 +150,41 @@ define(function(require) {
...
+     * If shouldShowDateInTitle is set to true, it takes the formatted Date.now() as additional influencer
+     * for the approximation of the needed number of lines.
+     * @param  {String}  text  Text which shall be tested for the necessary number of lines
+     * @param  {Number}  fontSize  Fontsize to use for the heuristic
+     * @param  {Number}  maxLength  Maximal length per line
+     * @return  {Number}  approximateLineNumber  Approximative number of lines needed to display the title
+     * @private
+     */
+    const getApproximateNumberOfLines = function(title, fontSize, maxLength) {
+        const words = title.split(/\s+/).reverse();
```
DalerAsrorov, line 164: is `title` always going to be defined?

src/charts/helpers/text.js
```diff
@@ -150,9 +150,41 @@ define(function(require) {
...
+
+        var word,
+            line = [],
+            approximateLineNumber = 1;
+
+        while (word = words.pop()) {
```
DalerAsrorov, line 170: Why not use a simpler for loop here?
```js
for (word in words) {
// stuff
}
```

src/charts/tooltip.js
```diff
@@ -401,15 +406,45 @@ define(function(require){
...
+         */
+        function updateTooltipTitleYPosition() {
+            const approximateTitle = getTooltipTitle(Date.now());
+            const approximateNumberOfTitleLines = getApproximateNumberOfLines(
+                approximateTitle,
+                16,
+                tooltipMaxTitleLength
+            );
+
+            if(approximateNumberOfTitleLines > 1) {
```
DalerAsrorov, line 421: a little bit of spacing needed here: 
```js
if (approximateNumberOfTitleLines > 1) {
...
```

src/charts/tooltip.js
```diff
@@ -401,15 +406,45 @@ define(function(require){
...
+        }
+
         /**
          * Updates value of tooltipTitle with the data meaning and the date
          * @param  {Object} dataPoint Point of data to use as source
          * @return void
          * @private
          */
         function updateTitle(dataPoint) {
+            const tTitle = getTooltipTitle(dataPoint[dateLabel]);
```
DalerAsrorov, line 433: What does `tTitle` represent? I think a bit more descriptive name for this would be helpful for the reader. Is it `textTitle`?

src/charts/helpers/text.js
ajdani: Would make sense, as it's an input without a default value. Doesn't make sense to use this function without `title`.

src/charts/helpers/text.js
ajdani: Yes, was sticking to the code-base. But changed it.

src/charts/tooltip.js
ajdani: Tbh I don't know what it stands for, just sticked to the naming which was there before - so I did not introduce this variable. But it makes sense to leave it a little bit cleaner as it was before, changed it to `textTitle`, because most likely that's the meaning.

ajdani:
Thanks for the feedback @DalerAsrorov !

Went over it and refactored the stuff mentioned.

ajdani:
Ok, Travis-CI strikes back :)
Now we have the issue, that `yarn run test:travis` runs locally, but Travis-CI does not like my test. 


Maybe we can have a look on this together if the results of yarn run test:travis should be deterministic we should fix the behavior of this rather than my failing test. @Golodhros @DalerAsrorov

I'm running node, npm and yarn in newer version as the CI - do you think this may cause the differences? Maybe different resolved dependencies...?
```
node --version
v12.13.0
npm --version
6.12.0
yarn --version
1.19.1
```

src/charts/helpers/text.js
```diff
@@ -150,9 +150,40 @@ define(function(require) {
...
+     * Heuristic which gets the number of lines needed to display the title of the tooltip
+     * If shouldShowDateInTitle is set to true, it takes the formatted Date.now() as additional influencer
+     * for the approximation of the needed number of lines.
+     * @param  {String}  text  Text which shall be tested for the necessary number of lines
+     * @param  {Number}  fontSize  Fontsize to use for the heuristic
+     * @param  {Number}  maxLength  Maximal length per line
+     * @return  {Number}  approximateLineNumber  Approximative number of lines needed to display the title
+     * @private
+     */
+    const getApproximateNumberOfLines = function(title, fontSize, maxLength) {
```
Golodhros, line 163: Could you set the default fontsize to 'defaultTextSize' like we do on the other helpers on this file?

Thank!

src/charts/helpers/text.js
```diff
@@ -150,9 +150,40 @@ define(function(require) {
...
+        const words = title.split(/\s+/).reverse();
+
+        let line = [],
+            approximateLineNumber = 1;
+
+        for(let word of words) {
+            line.push(word);
+
+            const textWidth = getTextWidth(line.join(' '), fontSize, 'Karla, sans-serif');
```
Golodhros, line 172: We probably want to add to the signature fontFace = defaultFontFace, like we do in the previous function and use the fontFace here

Golodhros:
Interesting, I actually get an error on your test when running the test:travis script on my machine.

This is what I have:
node 12.2.0
npm 6.9.0
yarn 1.19.2

ajdani:
Thanks. I will have a look on the failing test again, last time I could not reproduce it - but will check out your versions and fix it when I find the time for it!

ajdani:
Hi @Golodhros !

Was on this topic again, the test *does not* fail on my Windows machine, but *it fails* on my Ubuntu machine - the only difference seems to be the version of PhantomJS:
`PhantomJS 2.1.1 (Linux 0.0.0)`
vs.
`PhantomJS 2.1.1 (Windows 8.0.0)`

I would suggest to change the test to this, or even remove the check for the text of the tooltip in there: 
`it('should be displayed in two rows', () => {
                        // the space between 'Tooltip' and 'title' dissappears because of the text wrap
                        const expectedTitle = 'SuperlongandexcitingTooltiptitle-Aug05,2015';
                        const expectedDividerYPosition = 48;
                        let actualTitle, actualDividerY1Position, actualDividerY2Position;

                        tooltipChart.dateFormat(tooltipChart.axisTimeCombinations.DAY_MONTH);
                        tooltipChart.update({
                            date: '2015-08-05T07:00:00.000Z',
                            topics: []
                        }, topicColorMap, 0);

                        actualTitle = containerFixture.select('.britechart-tooltip')
                            .selectAll('.tooltip-title')
                            .text();

                        actualDividerY1Position = containerFixture.select('.britechart-tooltip')
                            .select('.tooltip-divider').attr('y1');

                        actualDividerY2Position = containerFixture.select('.britechart-tooltip')
                            .select('.tooltip-divider').attr('y2');

                        expect(actualTitle.split(" ").join("")).toEqual(expectedTitle);
                        expect(parseInt(actualDividerY1Position)).toEqual(expectedDividerYPosition);
                        expect(parseInt(actualDividerY2Position)).toEqual(expectedDividerYPosition);
                    });
`

Golodhros:
Hi @ajdani, thanks for following up!

Go for it! You can do whatever you feel necessary to make that test pass.

For the record, we are probably going to get rid of PhantomJS, as it doesn't work great with Github actions. Not sure when, but soon!

Thanks for your contribution!

Golodhros:
@all-contributors please add @ajdani for bug, code, maintenance

Golodhros:
Merged as https://github.com/britecharts/britecharts/pull/849

PR #828, opened 2020-08-04
Use fill:none; styling directly in SVG instead of external CSS file

ajdani:
I moved an important styling to the generated SVG directly instead of having it injected by the external CSS file describing several styles of the chart.

## Motivation and Context
When you currently extract the SVG of the chart and use Inkscape to generate a PDF out of it, the CSS styles are not loaded, therefore the exported PDF does not look nice (because of the `fill:none;` missing:
image

After my changes the generated PDF would look like this:
image

## How Has This Been Tested?
I did not change any tests nor added tests for it.

## Types of changes

## Checklist:
- [x] No linting issues
- [x] Build is successful
- [x] Code follows the [API Guidelines](http://britecharts.github.io/britecharts/topics-index.html#toc5__anchor)

### DIFF
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -882,7 +882,8 @@ define(function(require){
                   .attr('x2', chartWidth)
                   .attr('y1', (d) => yScale(d))
                   .attr('y2', (d) => yScale(d))
-                  .attr('stroke', (d) => getColor(d));
+                  .attr('stroke', (d) => getColor(d))
+                  .attr('fill', 'none');
 
             // draw the annotations right above the line at the right end of the chart
             for(let line of customLines) {
--- a/src/charts/line.js
+++ b/src/charts/line.js
@@ -791,7 +791,8 @@ define(function(require){
                 .attr('d', ({dates}) => topicLine(dates))
                 .style('stroke', (d) => (
                     dataByTopic.length === 1 ? `url(#${lineGradientId})` : getLineColor(d)
-                ));
+                ))
+                .style('fill', 'none');
 
             lines
                 .exit()
--- a/src/charts/sparkline.js
+++ b/src/charts/sparkline.js
@@ -327,7 +327,8 @@ define(function(require){
                 .attr('class', 'line')
                 .attr('stroke', `url(#${lineGradientId})`)
                 .attr('d', topLine)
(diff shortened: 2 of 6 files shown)


Golodhros:
Hi @ajdani, thanks for this PR!

This is interesting, just by replacing the fill:none the change is great. It still doesn't look awesome, I guess you are OK with that?

In the past, we have been using the export command to get the rendered charts as images, have you tried that? I have also heard about other people using puppeteer to simply render and capture screenshots of their charts to use on emails or PDFs, just FYI.

I would be OK with this change, but are these three charts the only ones that need it? Probably the rest do something similar? Just for completion sake, I would love to do it on all or none!

ajdani:
It does not look awesome, as many other stylings are missing, but at least okayish - I'm not sure if it makes sense to add all css styles into the `style` attribute of the SVG themselfes as it makes the code not really readable. I expect, that you mean the chart would look awesome if it's exactly the same as in the web?

Never tried the export functionality, if this works better than puppeteer it's worth a try, I'll investigate this. I tried it only with puppeteer to take a screenshot of a chart and it does not work out really well. And additionally you lose the advantage of a scalable vector graphic when you convert it into a pixel graphic.


I checked all occurences of `fill: none;` in the CSS and found nothing for the other chart-types and these were the only occurences. For e.g. donut-chart there should not be any problem. So I think it's already fixed for all charts and should work.

Golodhros:
Yep, I wouldn't go the way of adding all styles on the JS. It was something I thought before but discarded at least for now. Yeah, I was meaning to have all the styles to look awesome.

Oh, I see. Nice that you already tried puppeteer. The export function produces a PNG, so you would lose the SVG sharpness too. However, maybe you are interested in the logic on https://github.com/britecharts/britecharts/blob/master/src/charts/helpers/style.js, as it takes all the CSS styles and attaches them to every single HTML element. 

Check this option out and let me know, maybe you find a way of reusing that logic for your use case!

ajdani:
Thanks for merging @Golodhros !

And thanks for the hints for the style helper, I'm investigating on this already, I think that will fit nicely into our process and ensure that the SVGs converted to PDF will look nice. Unfortunately not done yet with it. :-(

Golodhros:
No problem!
This is still useful, thanks!

PR #877, opened 2020-11-06
chore: Deprecations

Golodhros:
## Description
Marks label methods as deprecated
Marks step chart as deprecated

## Motivation and Context
Communication with users

## How Has This Been Tested?
Manually

## Types of changes

-   [x] Refactor (changes the way we code something without changing its functionality)

## Checklist:

-   [ ] No linting issues
-   [ ] Build is successful

### DIFF
--- a/src/charts/bar.js
+++ b/src/charts/bar.js
@@ -14,6 +14,7 @@ import colorHelper from './helpers/color';
 import { bar as barChartLoadingMarkup } from './helpers/load';
 import { uniqueId } from './helpers/number';
 import { setDefaultLocale } from './helpers/locale';
+import { dataKeyDeprecationMessage } from './helpers/project';
 
 const PERCENTAGE_FORMAT = '%';
 const NUMBER_FORMAT = ',f';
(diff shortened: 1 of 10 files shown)


src/charts/bar.js
```diff
@@ -1122,12 +1123,17 @@ export default function module() {
...
      * @public
+     * @deprecated
      */
     exports.nameLabel = function (_x) {
         if (!arguments.length) {
             return nameLabel;
         }
         nameLabel = _x;
+        // eslint-disable-next-line no-console
+        console.log(
```
ImADrafter, line 1134: I think we could move this console.logs to a helper function. Its string could be taken by parameter, so we could:

- Remove `eslint-disable` from several lines and just have it on the helper.
- Make the code more dry.

Also, wouldn't it be more appropriate to use `console.warn` or `console.info` instead of log ? What do you think @Golodhros ?

Golodhros: Yeah, after the first three I was thinking the same!

Good call, console.warn will fit better!

ImADrafter (approved):

src/charts/bar.js
ImADrafter: :+1: :+1: :+1:

PR #879, opened 2020-11-10
fix: Fixes color mapping issues by adding the ColorMap property 

Golodhros:
## Description
Adds the colorMap property to all charts that need it
Updates types
Makes 'isAnimated' false by default in all charts
Removes uses of dataLabel, valueLabel, and other data labeling functions

## Motivation and Context
Fixes https://github.com/britecharts/britecharts/issues/431 and adds new possibilities to color configurations.
Part of https://github.com/britecharts/britecharts/issues/870

## How Has This Been Tested?
Added tests

## Types of changes

## Checklist:

-   [ ] No linting issues
-   [ ] Build is successful


demos/src/demo-stacked-bar.js
```diff
@@ -35,9 +35,6 @@ function createStackedBarChartWithTooltip(optionalColorSchema) {
             .width(containerWidth)
             .grid('horizontal')
             .isAnimated(true)
-            .stackLabel('stack')
```
DalerAsrorov, line 38: How does this change the rendering of the chart?

src/charts/bar.js
```diff
@@ -326,8 +325,17 @@ export default function module() {
...
-        colorMap = (item) =>
-            colorList.filter(({ name }) => name === item)[0].color;
+        nameToColorMap =
+            nameToColorMap ||
+            data
+                .map((d) => d)
+                .reduce(
+                    (acc, { name }, i) => ({
+                        ...acc,
+                        [name]: colorSchema[i % colorSchema.length],
```
DalerAsrorov, line 335: what does the modulo operator do in this case?

src/typings/charts/line-chart.d.ts
```diff
@@ -1,109 +1,113 @@
...
+    TopicName = 'topicName',
+    Value = 'value',
 }
 
 export type LineChartDataShape = {
-  [LineChartKeys.Date]: string;
-  [LineChartKeys.Name]: string;
-  [LineChartKeys.TopicName]: string;
-  [LineChartKeys.Value]: number;
+    [LineChartKeys.Date]: string;
```
DalerAsrorov, line 19: is the formatting still preserved? I assume it follows the eslint rules or Prettier configs we set

DalerAsrorov (approved):
few comments but looks great!

demos/src/demo-stacked-bar.js
Golodhros: It doesn't!

src/charts/bar.js
Golodhros: It applies one color from the color schema to each category, starting again when it reaches the max number of elements, so for a colorSchema of 2 colors:
- first element, i = 0, color colorSchema[0]
- second element, i = 1, color colorSchema[1]
- third element, i = 2, color colorSchema[0]
and so on

src/typings/charts/line-chart.d.ts
Golodhros: Prettier, yes.
We only lose the `.append` thing, but that gave more work than what was worth it!

PR #888, opened 2020-11-16
feat: Brush fixed window

Golodhros:
## Description
Allows users to fix the window of the brush chart, given an already set 'dateRange'

## Motivation and Context
Allow for fixed brush window

## How Has This Been Tested?
Added test
Tested manually on demo

## Types of changes

## Checklist:

-   [ ] No linting issues
-   [ ] Build is successful

### DIFF
--- a/src/charts/brush.js
+++ b/src/charts/brush.js
@@ -86,6 +86,7 @@ export default function module() {
         dateLabel = 'date',
         valueLabel = 'value',
         dateRange = [null, null],
+        isLocked = false,
         chartWidth,
         chartHeight,
         xScale,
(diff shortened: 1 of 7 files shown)


src/charts/brush.js
```diff
@@ -441,12 +442,39 @@ export default function module() {
...
 
-        dispatcher.call('customBrushStart', this, selection.map(xScale.invert));
+        if (isLocked) {
+            const lockedSelectionSize = Math.floor(
+                xScale(new Date(dateRange[1])) - xScale(new Date(dateRange[0]))
+            );
+            const selectedRange = Math.floor(selection[1] - selection[0]);
+
+            if (
+                selectedRange < lockedSelectionSize ||
```
DalerAsrorov, line 458: Wouldn't be simpler to do:
```js
selectedRange !== lockedSelectionSize
```

src/charts/brush.js
```diff
@@ -441,12 +442,39 @@ export default function module() {
...
+                newSelection = [
+                    Math.floor(selection[0]),
+                    Math.floor(selection[0]) + lockedSelectionSize,
+                ];
+                brush.move(chartBrush, newSelection);
+            } else {
+                newSelection = selection;
+            }
+        } else {
+            newSelection = selection;
```
DalerAsrorov, line 471: I see that:
```
 newSelection = selection;
```
assignment gets repeated. Maybe have a single case accounting for both instances?

src/charts/brush.js
Golodhros: True! Will change!!

src/charts/brush.js
Golodhros: Yeah, I thought about this. However, then we would need to calculate lockedSelectionSize and selectedRange for all cases, and this event triggers A LOT. That's why I chose this not so-elegant looking way.

I think the main issue here is that the event should probably be debounced!

PR #890, opened 2020-11-22
feat: turn exportChart into Promise

DalerAsrorov:
## Description

This PR turns `exportChart` into a promise that resolves or fails based on the success/failure of the chart image download process. This is going to be the only function that doesn't return the chart module but rather the promise that resolves if download happened or failed.

Would definitely need a PR review from @gagoar because it was his idea. Thanks a lot!

The user of the API will have the following form:
```js
donutChart.exportChart('donut.png', 'Britecharts Donut Chart')
  // image was downloaded succesfully
  .then(() => ...)
  // catch if there was an issue during download
  .catch(() => ...);
```

## Motivation and Context

https://github.com/britecharts/britecharts/issues/858

## How Has This Been Tested?

Revised tests, may add the `loadImage` test of its own if needed/asked.

## Types of changes

-   [x] Refactor (changes the way we code something without changing its functionality)
-   [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

-   [x] Latest master code has been merged into this branch
-   [x] No commented out code (if required, place // TODO above with explanation)
-   [x] No linting issues
-   [x] Build is successful
-   [x] Code follows the [API Guidelines](http://britecharts.github.io/britecharts/topics-index.html#toc5__anchor)
-   [x] Updated the documentation
-   [x] Added tests to cover changes
-   [x] All new and existing tests passed

### DIFF
--- a/src/charts/helpers/export.js
+++ b/src/charts/helpers/export.js
@@ -7,6 +7,8 @@ const isBrowser = typeof window !== 'undefined';
 const isIE = navigator.msSaveOrOpenBlob;
 const IE_ERROR_MSG =
     'Sorry, this feature is not available for IE. If you require this to work, check this issue https://github.com/eventbrite/britecharts/pull/652';
+const IMAGE_LOAD_DOWNLOAD_ERROR =
+    'Sorry, there was an issue with the chart download.';
 
 let encoder = isBrowser && window.btoa;
 
@@ -54,13 +56,21 @@ export function exportChart(d3svg, filename, title) {
         return false;
     }
 
-    let img = createImage(convertSvgToHtml.call(this, d3svg, title));
+    return loadImage(convertSvgToHtml.call(this, d3svg, title))
+        .then((img) => {
+            return {
+                canvas: createCanvas(this.width(), this.height()),
+                img,
+            };
+        })
+        .then(({ canvas, img }) => handleImageLoad.call(img, canvas, filename))
+        .then(() => true)
+        .catch(() => {
+            // eslint-disable-next-line no-console
+            console.error(IMAGE_LOAD_DOWNLOAD_ERROR);
 
-    img.onload = handleImageLoad.bind(
-        img,
-        createCanvas(this.width(), this.height()),
-        filename
-    );
+            return false;
+        });
 }
 
 /**
@@ -116,11 +126,15 @@ function createCanvas(width, height) {
 /**
  * Create Image
  * @param  {string} svgHtml string representation of svg el
+ * @param  {Function} callback function to prepare image for loading
  * @return {object}  TYPE element <img>, src points at svg
  */
-export function createImage(svgHtml) {
+function createImage(svgHtml, callback) {
     let img = new Image();
 
+    if (callback) {
+        callback(img);
+    }
     img.src = `${config.imageSourceBase}${b64EncodeUnicode(svgHtml)}`;
 
     return img;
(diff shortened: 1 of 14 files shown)


src/charts/helpers/export.js
```diff
@@ -54,13 +56,21 @@ export function exportChart(d3svg, filename, title) {
...
-    );
+    return loadImage(convertSvgToHtml.call(this, d3svg, title))
+        .then((img) => {
+            return {
+                canvas: createCanvas(this.width(), this.height()),
+                img,
+            };
+        })
+        .then(({ canvas, img }) => handleImageLoad.call(img, canvas, filename))
+        .then(() => true)
```
Golodhros, line 67: Why do we need this? is for defaulting to success? Not sure if I understand it.

DalerAsrorov: Great question. TBH we do not need this. I will remove it and let the user decide what the user wants to do. What do you think?

gagoar: maybe is the same question that @Golodhros asked, but a `Promise.resolve` or a `Promise.reject` should be the replacement for a true-false behavior.

src/charts/helpers/export.js
```diff
@@ -54,13 +56,21 @@ export function exportChart(d3svg, filename, title) {
...
+        .catch(() => {
```
gagoar, line 68: I will rethrow the error right here, so the user can do something else with it.

src/charts/helpers/export.js
```diff
@@ -116,11 +126,15 @@ function createCanvas(width, height) {
...
  * Create Image
  * @param  {string} svgHtml string representation of svg el
+ * @param  {Function} callback function to prepare image for loading
  * @return {object}  TYPE element <img>, src points at svg
  */
-export function createImage(svgHtml) {
+function createImage(svgHtml, callback) {
     let img = new Image();
 
+    if (callback) {
```
gagoar, line 135: What do you think about adding the check for the callback to be a function so we can error out with a helpful  message?

src/charts/helpers/export.js
DalerAsrorov: @gagoar great point. I believe the Promise constructor is already handling the `Promise.resolve` and `Promise.reject`. So if I remove that last `true/false` thing then the user will already bee able to have a `.then` callback if it's successful and `.catch` if it's a `Promise.reject`. Thanks for feedback!

src/charts/helpers/export.js
DalerAsrorov: This is a great suggestion. Will check do now.

Golodhros (approved):
LGTM

src/charts/helpers/export.js
```diff
@@ -116,11 +119,18 @@ function createCanvas(width, height) {
...
+        if (typeof callback !== 'function') {
+            throw new Error('The passed in callback should be a function');
```
gagoar, line 130:
```suggestion
            throw new Error(`The callback provided should be a function, we got a ${typeof callback} instead`);
```

src/charts/helpers/export.js
```diff
@@ -7,6 +7,8 @@ const isBrowser = typeof window !== 'undefined';
 const isIE = navigator.msSaveOrOpenBlob;
 const IE_ERROR_MSG =
     'Sorry, this feature is not available for IE. If you require this to work, check this issue https://github.com/eventbrite/britecharts/pull/652';
+const IMAGE_LOAD_DOWNLOAD_ERROR =
+    'Sorry, there was an issue with the chart download.';
```
gagoar, line 11:
```suggestion
    'Sorry, there was an issue downloading the chart.';
```

gagoar (approved):
some silly copy suggestions. 

Looks great!

PR #895, opened 2020-12-05
Fix: Fixes issue where bullet won't get the right range

Golodhros:
## Description
Fixes problem with bullet not getting ranges properly

## Motivation and Context
Bug fix

## How Has This Been Tested?
Ran tests
Checked manually with extended example

Tutorial__Bullet_Chart

## Types of changes

## Checklist:

-   [ ] No linting issues
-   [ ] Build is successful

### DIFF
--- a/test/fixtures/bulletChartDataBuilder.js
+++ b/test/fixtures/bulletChartDataBuilder.js
@@ -35,6 +35,50 @@ define(function(require) {
             return new this.Klass(attributes);
         };
 
+        this.withManyCPUData = function () {
+            var attributes = _.extend({}, this.config, {
+                data: [
+                    {
+                        title: 'CPU 1',
+                        subtitle: 'GHz',
+                        ranges: [1500, 2100, 3500],
+                        measures: [1800, 2200],
+                        markers: [2600, 3000],
+                    },
+                    {
+                        title: 'CPU 2',
+                        subtitle: 'GHz',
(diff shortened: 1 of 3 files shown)


test/fixtures/bulletChartDataBuilder.js
```diff
@@ -35,6 +35,50 @@ define(function(require) {
             return new this.Klass(attributes);
         };
 
+        this.withManyCPUData = function () {
```
DalerAsrorov, line 38: for consistency sake, let's have a json file for this dataset as well

DalerAsrorov (approved):

test/fixtures/bulletChartDataBuilder.js
Golodhros: I am actually thinking of dumping the JSON files.

They add a level of indirection that is a pain, especially when dealing with tests and examples. It is true that there are a couple of datasets (line and area chart) that are large, but most of the time is not a big deal and our IDE's collapsing features help a lot.

what do you think?

DalerAsrorov: Sounds good to me. If we were using TypeScript dataset types could be highlighted if we were to place a wrong key or value type, so I would go with not using json files as well.

Golodhros: Cool, I will go ahead and refactor that on the RC

PR #905, opened 2021-03-12
fix: Fixes bullet chart issues

Golodhros:
## Description
Fixes issues with code and tests of the bullet chart

## Motivation and Context
Fixing stuff

## How Has This Been Tested?
Manually and through tests

## Types of changes

### DIFF
--- a/src/charts/bullet.js
+++ b/src/charts/bullet.js
@@ -14,13 +14,15 @@ import colorHelper from './helpers/color';
  * @property {Number[]} measures    Range that encodes the performance measure
  * @property {Number[]} markers     Marker lines that encode the comparative measure
  * @property {String}   [title]     String that sets identification for the measure
- * @property {String} [subtitle]  String that provides more details on measure identification
+ * @property {String}   [subtitle]  String that provides more details on measure identification
  *
  * @example
  * {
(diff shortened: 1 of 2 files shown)


src/charts/bullet.js
```diff
@@ -309,7 +306,7 @@ export default function module() {
                 .enter()
                 .append('line')
                 .attr('class', 'marker-line')
-                .attr('stroke', measureColor)
+                .attr('stroke', markerColor)
```
DalerAsrorov, line 309: Are we sure we need to use marketColot instead? Just want to make sure that nothing looks off when we switch colors in demo

src/charts/bullet.js
```diff
@@ -325,6 +322,10 @@ export default function module() {
      * @private
      */
     function drawHorizontalExtendedLine() {
+        if (baseLine) {
+            baseLine.remove();
```
DalerAsrorov, line 326: Thanks I forgot to do all that last time

DalerAsrorov (approved):
Lgtm

src/charts/bullet.js
Golodhros: Yep, that way we can configure that color too. I think it was intended, but we missed it somehow.

PR #971, opened 2021-11-15
Add hasMinimumValueScale parameter to autoscale linechart y-axis

alexisbcc:
## Description

For Line charts a parameter `hasMinimumValueScale` to adjust the Y axis scale in the case of only positive and non-zero values was added, the parameter is `false` by default.

## Motivation and Context

Issue: https://github.com/britecharts/britecharts/issues/866

## How Has This Been Tested?

When the data set contains only positive values greater that zero the y-axis should have a minimum value 
positive and non-zero, and the maximum closer or equal the max value of the data set.
image

## Types of changes

## Checklist:

-   [x] No linting issues
-   [x] Build is successful
-   [x] Code follows the [API Guidelines](http://britecharts.github.io/britecharts/topics-index.html#toc5__anchor)
-   [x] Added tests to cover changes
-   [x] All new and existing tests passed


src/charts/line/line.js
```diff
@@ -1785,6 +1786,24 @@ export default function module() {
...
+    /**
+     * Gets or Sets the hasMinimumValueScale property of the chart, making yAxix bottom value
+     * to adjust to the minimum dataset value.
+     * By default this is 'false'
+     *
+     * @param  {Boolean} _x Desired minimum value flag
+     * @return { hasMinimumValueScale | module} Current hasMinimumValueScale flag or Chart module
+     * @public
+     */
+    exports.hasMinimumValueScale = function (_x) {
```
Golodhros, line 1798: Could you order this alphabetically? It makes it easier to maintain later.
Thanks!

src/charts/line/lineDataOnlyNonzeroPositiveValues.json
```diff
@@ -0,0 +1,42 @@
+{
+	"dataByTopic": [
```
Golodhros, line 2: We are moving out of this kind of data shape and into a more flat structure, check out the other .json files in this folder to see it in practice.

It would be great if you could use that shape.

Golodhros (commented):
Looking good!
Some minor comments.

src/charts/line/line.js
alexisbcc: Should I move the initial declaration too? it seems that those variables are not alphabetically ordered.
What is the order criterion of these variables?

Golodhros: Nope, those are not ordered for now. It started one way, then changed... don't worry about those.

alexisbcc: ok, I think it should be good now, let me know if this is correct.

Golodhros (approved):
LGTM

Thanks a lot!

Golodhros:
@all-contributors add @alexisbcc for 💻

Golodhros:
@all-contributors add @alexisbcc for code

PR #976, opened 2022-01-10
typings fixes

shayh:


Golodhros:
Thanks a lot for these fixes @shayh!!

Golodhros:
It is live on 3.0.0-alpha-6.1.3!

PR #994, opened 2022-03-02
docs: updates changelog and workflow update

Golodhros:
## Description
Updates changelog 
Updates npm-publish workflow

## Motivation and Context
Doc updates

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -8,25 +8,25 @@ on:
     types: [created]
 
 jobs:
+  getNodeVersion:
+    uses: calyptia/.github/.github/workflows/get_node_version.yml@main
+
   build:
     runs-on: ubuntu-latest
+    needs: getNodeVersion
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -8,25 +8,25 @@ on:
...
-      - name: Install and Test
-        uses: actions/setup-node@v1
+      - name: Setting node version to ${{ needs.getNodeVersion.outputs.node_version }}
+        uses: actions/setup-node@v2
         with:
-          node-version: 12.2.0
-      - uses: borales/actions-yarn@v2.0.2
+          node-version: '${{ needs.getNodeVersion.outputs.node_version }}'
+      - name: Install and Test
+        uses: borales/actions-yarn@v2.0.2
```
gagoar, line 27: I'm curious here, there's no need for you to use an action to do yarn. instead, you should be able to copy the cache and install what we did in the previews iteration.

PR #1006, opened 2022-05-23
Linting js

gagoar:
## Description
We need linting for mono repo working. as requested over at #834 

Some issues with `stylelint`: 

[Here](https://github.com/stylelint/stylelint/issues/5766) it seems to be suggested that some version's combination will do the trick but I tried and it does not work. 

There are several discussions referring to commenting styles that are breaking `stylelint`, `//` breaks it, and `/*` doesn't but I didn't get to test this. 

## Motivation and Context
I feel the need, the need for lint!

## How Has This Been Tested?
I'm running it on my computer, but this PR will hopefully run with the lint workflow. 

## Types of changes

## Checklist:

-   [ ] No linting issues
-   [ ] Build is successful


gagoar:
Have you changed anything? I don't see the workflows running 🤔

gagoar:
> Have you changed anything? I don't see the workflows running 🤔

@Golodhros ^

Golodhros:
> Have you changed anything? I don't see the workflows running 🤔

Nope!
But I see a "Verified" tag on your commits, so I think they are running, right?

package.json
```diff
@@ -121,17 +121,39 @@
...
+    "settings": {
+      "react": {
+        "version": "detect"
+      }
+    },
+    "extends": [
+      "plugin:react/recommended",
+      "plugin:react/jsx-runtime",
+      "plugin:testing-library/react"
+    ],
```
Golodhros, line 133: Oh, I was surprised to see this here as these are React specific

packages/core/src/charts/donut/donut.spec.js
```diff
@@ -666,8 +666,8 @@ donutDataSets.forEach((datasetName) => {
 
                     donutChart.on('customClick', callback);
                     firstSlice.dispatch('click');
-                    expect(callback.mock.calls.length).toBe(1);
-                    expect(callback.mock.calls[0].length).toBe(3);
+                    expect(callback.mock.calls).toHaveLength(1);
```
Golodhros, line 669: I will probably review this settings, as I do prefer not to use matchers as much as possible, relying on regular JS and a .toEqual or .toBe for everything.

Golodhros (approved):

package.json
gagoar: It is needed given that we have JSX in the tests.

packages/core/src/charts/donut/donut.spec.js
gagoar: Some of these changes came from jest recommended rules. I can revert. But it is the recommended way

package.json
Golodhros: Yeah, but only on the /react package, right?

packages/core/src/charts/donut/donut.spec.js
Golodhros: No prob, I will polish it later on. Not a big deal.

package.json
gagoar: Yea but liting shares a signle configuation.

PR #1008, opened 2022-07-04
docs: Updates docs homepage, edit markdown files, adds package docs to site

Golodhros:
## Description
Many docs updates and tweaks
Links are mostly broken

## Motivation and Context
Finishing the docs

## How Has This Been Tested?
In dev machine


gagoar (approved):
This is too much for me to review, so I just checked things I thought were important. 

💪🏻

packages/core/src/styles/britecharts.scss
```diff
@@ -1,24 +1,24 @@
 // Base Styles
-@import 'base/_typography';
+@import "base/_typography";
 
 // Helpers
-@import 'helpers/_variables';
-@import 'helpers/loadingStates';
+@import "helpers/_variables";
+@import "helpers/loadingStates";
```
gagoar, line 6: why changing to double quotes?

Golodhros: Not sure why (I think a bug related to the three stylelint recommendation sets we were loading) I had to change slightly the stylelint settings. This was part of it. 
I really don´t care as long as they get fixed automatically!

PR #1018, opened 2023-02-09
chore: testing chromatic setup

Golodhros:
## Description
Fixes storybook composed in production
Publishes Core's storybook
Publishes React's storybook

## Motivation and Context
Getting demos to work

## How Has This Been Tested?
Manually


packages/demos/package.json
```diff
@@ -8,11 +8,11 @@
...
   "scripts": {
     "build": "build-storybook -c .storybook -o dist",
+    "chromatic": "chromatic --project-token=$CHROMATIC_TOKEN",
     "clean": "rimraf dist",
     "demo": "yarn storybook",
     "prebuild": "yarn run clean",
     "start": "yarn storybook",
-    "storybook": "start-storybook -p 9999",
-    "chromatic": "chromatic --project-token=$CHROMATIC_TOKEN"
+    "storybook": "start-storybook -p 9999"
```
gagoar, line 16: do you mind using ports that are one next to each other instead?

gagoar (approved):
just a bit of organization around the ports you use. I would like it to see it one after the other instead of 1111 and 9999

packages/core/package.json
```diff
@@ -135,14 +136,16 @@
...
-    "demo": "start-storybook -p 1111",
+    "demo:build": "build-storybook -c .storybook -o dist/storybook",
+    "demo": "yarn run storybook",
     "lint:js": "eslint src/charts/ webpack.*.js --cache",
     "lint:styles": "stylelint './src/styles/**/*.scss'",
     "lint": "yarn run lint:js && yarn run lint:styles",
     "prebuild": "yarn run clean",
     "start": "yarn run demo",
-    "storybook:build": "build-storybook -c .storybook -o build",
+    "storybook": "start-storybook -p 1111",
```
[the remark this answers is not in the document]
Golodhros: Don't know why we choose those back in the day!

packages/demos/package.json
Golodhros: like 1001, 1002, 1003?

I guess it will work too.

Could we somehow parametrize this though, so we could use it on the storybook config?

Golodhros: OK, will do!

